PR #4, opened 2020-02-10
Generalize the interface

afilini:
take `IntoIterator` instead of a `Vec` and also implement `From<S>` for `Client` where `S` is a generic `Read + Write` type

Also refactor the whole "streams" part


RCasatta:
utack 276a03d56dfc2a5418e4411ad2e8488757c6adee

PR #5, opened 2020-02-11
Take a bool for SSL domain validation

afilini:


src/client.rs
```diff
@@ -113,20 +135,26 @@ impl Client<ElectrumPlaintextStream> {
...
+    /// `validate_domain` is `true`, validate the server's certificate.
+    pub fn new_ssl<A: ToSocketAddrsDomain>(
+        socket_addr: A,
+        validate_domain: bool,
+    ) -> Result<Self, Error> {
         let mut builder =
             SslConnector::builder(SslMethod::tls()).map_err(Error::InvalidSslMethod)?;
         // TODO: support for certificate pinning
-        if domain.is_none() {
+        if !validate_domain {
```
RCasatta: nit: inverting if branch to avoid the not

src/client.rs
```diff
@@ -49,6 +49,28 @@ macro_rules! impl_batch_call {
...
+pub trait ToSocketAddrsDomain: ToSocketAddrs {
+    /// Returns the domain, if present
+    fn domain(&self) -> Option<&str> {
+        None
+    }
+}
+
+impl ToSocketAddrsDomain for &str {
+    fn domain(&self) -> Option<&str> {
+        self.splitn(2, ':').into_iter().next()
```
RCasatta: nit: clippy says you can remove into_iter()

afilini: fixed in 78a8a7c

src/client.rs
afilini: fixed in 78a8a7c

RCasatta:
utack  68cd825

PR #7, opened 2020-02-19
Add estimate_fee to batch methods

RCasatta:


RCasatta:
Do we want a `Batch::new(calls)` method so that a library user could manually build some calls if they are not present?

jb55:
> Do we want a `Batch::new(calls)` method so that a library user could manually build some calls if they are not present?

that makes sense to me

jb55:
Tested ACK 5e1d23a6750a98f557cb8164bb502de2e581696a

afilini:
For some reason the CI didn't run, I re-pushed it on a different branch and now it has picked it up. As I'm writing this it's still running, but the formatting looks a bit odd to me, so I'm afraid it will fail because of it.

Could you `cargo fmt` and re-push?

RCasatta:
fixed formatting, but it looks CI is not starting

afilini:
Will look into why the CI doesn't work automatically, but this is [looking good](https://travis-ci.org/MagicalBitcoin/rust-electrum-client/builds/657101155). Merging now.

PR #8, opened 2020-02-24
Deser header

RCasatta:
~~include #7~~

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -107,6 +107,16 @@ where
     Ok(answer)
 }
 
+fn from_hex_header<'de, D>(deserializer: D) -> Result<block::BlockHeader, D::Error>
+where
+    D: de::Deserializer<'de>,
+{
+    use bitcoin::consensus::deserialize;
+
+    let vec: Vec<u8> = from_hex(deserializer)?;
+    deserialize(&vec).map_err(de::Error::custom)
+}
+
 /// Response to a [`script_get_history`](../client/struct.Client.html#method.script_get_history) request.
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -189,8 +200,8 @@ pub struct GetMerkleRes {
 pub struct HeaderNotification {
     /// New block height.
     pub height: usize,
-    #[serde(rename(serialize = "hex"))]
     /// Newly added header.
+    #[serde(rename(serialize = "hex", deserialize = "hex"), deserialize_with = "from_hex_header")]
```
afilini: if both `serialize` and `deserialize` are the same I guess you could use [`rename = "hex"`](https://serde.rs/field-attrs.html#rename)

src/types.rs
```diff
@@ -107,6 +107,17 @@ where
     Ok(answer)
 }
 
+fn from_hex_header<'de, D>(deserializer: D) -> Result<block::BlockHeader, D::Error>
```
afilini: i might be missing something here, but aren't those types like `BlockHeader` already de-serializable? why did you have to re-implement the serializer manually?

RCasatta: It just didn't work out of the box

src/types.rs
RCasatta: fixed

src/types.rs
RCasatta: it just doesn't work out of the box for me

RCasatta: maybe because BlockHeader is missing FromStr

Issue #9, opened 2020-03-09
block_headers_poll is blocking but it should not be


RCasatta:
block_headers_poll is blocking but it should not be, this test hangs

```
    #[test]
    fn test_poll() -> Result<(), Error> {
        let mut client = electrum_client::Client::new("kirsche.emzy.de:50001")?;
        let headers = client.block_headers_subscribe()?;
        println!("{:?}", headers);
        let block = client.block_headers_poll()?;

        Ok(())
    }
```

PR #10, opened 2020-03-09
[WIP] Hack: make `poll()` non-blocking when required

afilini:
This PR tries to fix issue #9 with a pretty weird hack: the idea here is that `BufReader::fill_buf()` is blocking, meaning that it will wait until it can pull some data from the underlying stream. I was expecting it to be non-blocking because I thought that internally it would poke the OS, see if there's something for us, and if not just exit. But actually, it makes sense for it to be blocking, since the trait it implements (`BufRead`) relies on the `Read` trait, which is also blocking. I guess the lesson here is that you can't really build non-blocking functions on top of blocking primitives.

So, going back to the hack: since we really want the server to send something to us (even something that we can drop), we will send a `server.ping` before trying to fill the buffer. Since the server will always reply to our pings, we will always find something in the buffer when we go there and try to read from it. Sometimes it's just the ping, sometimes there are also some notifications for us.

I guess the right way to fix this issue would be to move to full asynchronous I/O, but I'm not an expert and it needs more research from my side. Another option would be to try to find some library that allows us to query the kernel and see if there's something for us (see `(ioctl/FIONREAD)[https://www.freebsd.org/cgi/man.cgi?query=ioctl&sektion=2]`).

If we need a quick fix I would be fine merging this, considering that the library is still marked as `beta`. Meanwhile I can do more research and see if there's something lightweight that does what we need here.

Still WIP because the docs have not been updated. If the fix looks good to you @RCasatta I will update those and merge the PR.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -344,7 +344,11 @@ impl<S: Read + Write> Client<S> {
     ) -> Result<(), Error> {
         match method {
             "blockchain.headers.subscribe" => {
-                self.headers.push_back(serde_json::from_value(result)?)
+                if result.is_array() {
+                    for val in result.as_array().unwrap() {
+                        self.headers.push_back(serde_json::from_value(val.clone())?)
+                    }
+                }
             }
             "blockchain.scripthash.subscribe" => {
                 let unserialized: ScriptNotification = serde_json::from_value(result)?;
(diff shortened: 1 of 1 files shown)


RCasatta:
interesting work around, but kinda hacky, I don't need the poll method at the moment so your call if you want to merge

afilini:
Closing since #18 will implement this

PR #14, opened 2020-03-12
add batch_block_header and fix batch_transaction_get

RCasatta:


src/client.rs
```diff
@@ -595,14 +593,34 @@ impl<S: Read + Write> Client<S> {
...
+        I: IntoIterator<Item = usize>,
+    {
+        let headers_string: Result<Vec<String>, Error> =
+            impl_batch_call!(self, heights, block_header);
+        let mut headers = vec![];
+        for header_string in headers_string? {
+            let header: BlockHeader = deserialize(&Vec::<u8>::from_hex(&header_string)?)?;
+            headers.push(header)
+        }
+        Ok(headers)
```
afilini: since `Result` implements `FromIterator`, you can use `collect()` over a vec of results to get a result of vec (see [here](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect))

```suggestion
        headers_string?
            .into_iter()
            .map(|string| Ok(deserialize(&Vec::<u8>::from_hex(&string)?)?))
            .collect()
```

src/client.rs
```diff
@@ -595,14 +593,34 @@ impl<S: Read + Write> Client<S> {
...
         I: IntoIterator<Item = &'t Txid>,
     {
-        impl_batch_call!(self, txids, transaction_get)
+        let txs_string: Result<Vec<String>, Error> = impl_batch_call!(self, txids, transaction_get);
+        let mut txs = vec![];
+        for tx_string in txs_string? {
+            let tx: Transaction = deserialize(&Vec::<u8>::from_hex(&tx_string)?)?;
+            txs.push(tx)
+        }
+        Ok(txs)
```
afilini: same as below, you can use `collect()` over a vec of results

src/client.rs
RCasatta: nice, going to change

PR #17, opened 2020-03-24
add raw methods for transactions and headers

RCasatta:


afilini:
Looks good to me, the only nit i have is about the documentation: the names of the methods are pretty self explanatory, but I would add a couple of words to explain that the `raw` version returns the bytes, while the other one already deserializes them into structs. Maybe somebody who is not familiar with rust-bitcoin would think that the raw version returns the bytes and the other one some kind of JSON, but it's not actually like that, they both return essentially the same thing.

afilini:
Merged in 396ab8ea57de24ad3bf6238efb5838eae47a2701, the SHA changed because I amended your commit with `cargo fmt`.

Thanks!

PR #18, opened 2020-04-08
Move to async I/O with Tokio

afilini:


RCasatta:
ack e1dbfd7ceb612ae9a82d1e5782aa184be5a63f02

afilini:
Do you mean tokio as a dependency? because i don't think it can be removed in any way, unless you implement the logic both in a sync and async way. Currently the main logic is asynchronous, so you'll always need tokio for that, even if the exposed interface is synchronous.

RCasatta:
no I mean tokio as runtime, to use sync methods you still need to run the tokio runtime

RCasatta:
ack dcad619

PR #19, opened 2020-07-12
Multithreading

afilini:


src/client.rs
```diff
@@ -250,21 +262,119 @@ impl<S: Read + Write> Client<S> {
...
+                                // We have an id, but it's not our response. Notify the thread and
+                                // move on
+                                trace!(
+                                    "Reader thread {}, received response for {}",
+                                    req.id,
+                                    resp_id
+                                );
+
+                                let mut map = self.waiting_map.write().unwrap();
+                                if map.get(&resp_id).is_some() {
```
RCasatta, line 343: if let Some(sender) = map.remove(&resp_id) {
 sender.send(resp);
}

src/client.rs
```diff
@@ -250,21 +262,119 @@ impl<S: Read + Write> Client<S> {
...
+                                    "Reader thread {} received a response for its request",
+                                    req.id
+                                );
+
+                                // Remove ourselves from the "waiting map"
+                                let mut map = self.waiting_map.write().unwrap();
+                                map.remove(&req.id);
+
+                                // If the map is not empty, we select a random thread to become the
+                                // new reader thread. We close its channel to "wake it up"
```
RCasatta, line 320: maybe clearer to use an explicit enum as the value sent across the channel to receive data or be woke up, something like:
```
enum ValOrWake {
 Val(Value),
WakeUp,
}
```

src/client.rs
afilini: It was like you proposed initially, but then I realized I was removing something from the map while i was also borrowing it from the map and I decided to change it.

Maybe it's actually possible to get it to work, I'll try

src/client.rs
afilini: yeah that's a good idea!

RCasatta:
ack 335fd49eb52f23e54b37b41761dc22f2f8114f8e

PR #21, opened 2020-07-31
allow Client with openssl+proxy

RCasatta:
without these changes you can't for example run:

`cargo run --example ssl --no-default-features --features use-openssl --features proxy`

because `Client` is built only with with default features, but having the combination of SSL and proxy is enough for the client to run.

to avoid repeating the cfg before Client, impl Client and impl ElectrumApi the cfg has been moved at module level

it also change the Cargo toml to recall features instead of repeating deps for the default feature

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -11,8 +11,7 @@ use types::*;
 /// [`RawClient`](client/struct.RawClient.html) and provides a more user-friendly
 /// constructor that can choose the right backend based on the url prefix.
 ///
-/// **This is only available with the `default` features.**
-#[cfg(feature = "default")]
+/// **This is available only with the `default` features, or if `proxy` and one ssl implementation are enabled**
 pub enum Client {
     #[doc(hidden)]
     TCP(RawClient<ElectrumPlaintextStream>),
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -10,9 +10,6 @@ use types::*;
 /// Generalized Electrum client that supports multiple backends. This wraps
 /// [`RawClient`](client/struct.RawClient.html) and provides a more user-friendly
 /// constructor that can choose the right backend based on the url prefix.
-///
-/// **This is only available with the `default` features.**
```
afilini, line 14: I would keep a similar comment that maybe says "This is available only with the default features, or if `proxy` and one ssl implementation are enabled"

src/lib.rs
```diff
@@ -40,7 +40,14 @@ extern crate webpki_roots;
 
 mod api;
 mod batch;
+
+#[cfg(any(
+    feature = "default",
```
afilini, line 45: Not 100% sure about this, but I think this line can be removed, because `default` implies `proxy` and `use-rustls`, which should be covered by the lines below

src/client.rs
RCasatta: done

src/lib.rs
RCasatta: it looks it works without `default`, done

PR #25, opened 2020-11-07
Allow to initialize SSL clients with an existing TcpStream

shesek:
To make it easier to configure custom TcpStream options like timeouts, without duplicating the SSL stream initialization code.


afilini:
I hadn't realized this at first, but I think this could also be used to add support for SSL over SOCKS5. I think the socks library I'm using allows getting the `TcpStream` back once it's done with the "handshake", I guess with authentication and things like that. So in theory you could create a `TcpStream`, give that to the socks library to authenticate with the proxy, and then get it back and give it to openssl and do the ssl handshake and establish a secure connection.

I'll see if I can make that work, that would be a nice addition!

afilini:
Looks like this works! #26 

I didn't have to change anything here, so I'll merge this PR right now and then work a bit more on my own to add the proxy stuff everywhere and in the docs

shesek:
@afilini Awesome! :100:

PR #27, opened 2020-11-20
Config

RCasatta:
## Implements retry, timeout, socks credential and others fixes
    
Retry works by checking if an non-protocol error happened during the call, if so an attempt to take a write lock on the client is made and a new client attempted
    if both operationare succesfull the old_client is substituted with the new one.
    
Timeout is a single parameter for connect,read and write TCP timeouts, special handling is required for the following situation: cannot set a timeout with a proxy and cannot set timeout with multipl SocketAddrs
    
Since configurations parameter grow, a Config struct with a builder has been introduced
    
Some new errors variant has been introduced
    
Errors in reading input in the reader thread now warns other eventual threads with ChannelMessage::Error

Add github actions equivalent to travis cause travis is slow

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,9 +1,14 @@
 //! Electrum Client
 
+use std::sync::RwLock;
+
+use log::{info, warn};
+
 use bitcoin::{Script, Txid};
 
 use api::ElectrumApi;
 use batch::Batch;
+use config::Config;
 use raw_client::*;
 use types::*;
 
@@ -12,7 +17,7 @@ use types::*;
 /// constructor that can choose the right backend based on the url prefix.
 ///
 /// **This is available only with the `default` features, or if `proxy` and one ssl implementation are enabled**
-pub enum Client {
+pub enum ClientType {
     #[doc(hidden)]
     TCP(RawClient<ElectrumPlaintextStream>),
     #[doc(hidden)]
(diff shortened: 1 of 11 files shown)


src/client.rs
```diff
@@ -21,56 +26,124 @@ pub enum Client {
...
+                None => ClientType::TCP(RawClient::new(url.as_str(), config.timeout())?),
+                Some(socks5) => ClientType::Socks5(RawClient::new_proxy(url.as_str(), socks5)?),
+            })
         }
     }
 }
 
 impl Client {
-    /// Generic constructor that supports multiple backends and, optionally, a socks5 proxy.
+    /// Default constructor supporting multiple backend by providing a prefix
```
afilini, line 114: Typo: "... supporting multiple backends ..."

src/client.rs
```diff
@@ -21,56 +26,124 @@ pub enum Client {
...
+                    if errors.len() as u8 == $self.config.retry() {
+                        return Err(Error::AllAttemptsErrored(errors));
+                    }
+
+                    // Only one thread will try to recreate the client getting the write lock,
+                    // other eventual threads will get Err and will block at the beginning of
+                    // previous loop when trying to read()
+                    if let Ok(mut write_client) = $self.client_type.try_write() {
+                        loop {
+                            std::thread::sleep(std::time::Duration::from_secs(errors.len() as u64));
```
afilini, line 66: Do you think it's better to keep this linear or switch to an exponential backoff, potentially with a cap to a reasonable value like 16 seconds or something like that?

src/config.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn validate_domain(&self) -> bool {
+        self.validate_domain
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Config {
+            socks5: None,
+            timeout: None,
```
afilini, line 125: Do you know what's the default timeout? Is it reasonable or maybe is it better to override it with something else by default?

src/client.rs
RCasatta: what do you think of something like `(errors.len() * errors.len()).min(60)`

src/config.rs
RCasatta: at the moment it's not convenient in case of proxy to have a default value, because you would need to set it to None before setting the proxy

afilini: Right, makes sense

src/client.rs
afilini: I was thinking about `(1 << errors.len()).min(something)` but that should also work. Not sure what's the best really ahaha

Also not sure what's a good value for the cap: 60 seconds between two tries seems a bit high to me, maybe around 30 is better?

RCasatta: Ok, now `(1 << errors.len()).min(30)`

src/client.rs
RCasatta: fixed in f0e863f1cda111ec746ed2d95923db234e85d8fd

src/raw_client.rs
```diff
@@ -336,7 +398,13 @@ impl<S: Read + Write> RawClient<S> {
                 loop {
                     raw_resp.clear();
 
-                    reader.read_line(&mut raw_resp)?;
+                    if let Err(e) = reader.read_line(&mut raw_resp) {
+                        let error = Arc::new(e);
+                        for (_, s) in self.waiting_map.lock().unwrap().drain() {
+                            s.send(ChannelMessage::Error(error.clone()))
+                                .expect("Unable to send ChannelMessage::Error");
+                        }
```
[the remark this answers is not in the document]
RCasatta: yes, fixed in f0e863f1cda111ec746ed2d95923db234e85d8fd
I also changed the enum variant to `SharedIOError`

PR #33, opened 2021-01-20
Getting rid of lock().unwrap()

dr-orlovsky:
PR removes 16 (!) `unwraps()` which are unnecessary and were already covered by an error case.

This is quite important for the overall library stability

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -452,13 +452,7 @@ impl<S: Read + Write> RawClient<S> {
                     // might have already received a response for that id, but we don't know it
                     // yet. Exiting here forces the calling code to fallback to the sender-receiver
                     // method, and it should find a message there waiting for it.
-                    if self
-                        .waiting_map
-                        .lock()
-                        .unwrap()
-                        .get(&until_message)
-                        .is_none()
-                    {
+                    if self.waiting_map.lock()?.get(&until_message).is_none() {
                         return Err(Error::CouldntLockReader);
                     }
                 }
@@ -494,7 +488,7 @@ impl<S: Read + Write> RawClient<S> {
                             );
 
                             // Remove ourselves from the "waiting map"
-                            let mut map = self.waiting_map.lock().unwrap();
+                            let mut map = self.waiting_map.lock()?;
                             map.remove(&resp_id);
 
                             // If the map is not empty, we select a random thread to become the
@@ -512,8 +506,7 @@ impl<S: Read + Write> RawClient<S> {
                             // move on
                             trace!("Reader thread received response for {}", resp_id);
 
-                            if let Some(sender) = self.waiting_map.lock().unwrap().remove(&resp_id)
-                            {
+                            if let Some(sender) = self.waiting_map.lock()?.remove(&resp_id) {
                                 sender
                                     .send(ChannelMessage::Response(resp))
                                     .expect("Unable to send the response");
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -322,3 +324,21 @@ impl_error!(std::io::Error, IOError);
 impl_error!(serde_json::Error, JSON);
 impl_error!(bitcoin::hashes::hex::Error, Hex);
 impl_error!(bitcoin::consensus::encode::Error, Bitcoin);
+
+impl<T> From<std::sync::PoisonError<T>> for Error {
+    fn from(_: std::sync::PoisonError<T>) -> Self {
+        Error::CouldntLockReader
```
afilini, line 330: I wouldn't use this error here, because `CouldntLockReader` has a very specific meaning:

> Couldn't take a lock on the reader mutex. This means that there's already another reader thread running

If a `PoisonError` happens there's not much the caller can do to recover, it has to throw away the client and recreate it. I would add an error variant that reflects this better, and write some documentation to let users know what to do in case that error happens.

src/stream.rs
```diff
@@ -6,17 +6,26 @@ pub struct ClonableStream<T: Read + Write>(Arc<Mutex<T>>);
 
 impl<T: Read + Write> Read for ClonableStream<T> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        self.0.lock().unwrap().read(buf)
+        self.0
+            .lock()
+            .map_err(|_| io::Error::from(io::ErrorKind::BrokenPipe))?
```
afilini, line 11: I agree that `BrokenPipe` is the error variant that more closely maps to a `PoisonError`, but I would at least log what's happening here first before returning, to avoid confusing the user.

Same for `write()` and `flush()` below

src/raw_client.rs
```diff
@@ -472,15 +465,13 @@ impl<S: Read + Write> RawClient<S> {
...
+                            let mut map = self.waiting_map.lock()?;
                             map.remove(&resp_id);
 
                             // If the map is not empty, we select a random thread to become the
                             // new reader thread.
                             if let Some(sender) = map.values().next() {
-                                sender
-                                    .send(ChannelMessage::WakeUp)
-                                    .expect("Unable to WakeUp a different thread");
+                                sender.send(ChannelMessage::WakeUp)?;
```
afilini, line 474: If we are unable to wake up the thread I guess we should try with all the others before giving up, otherwise all the other threads that are still alive would remain hanging forever waiting for their response.

afilini (changes requested):

src/types.rs
dr-orlovsky: I am strongly against panics at runtime which happens not due to app design bugs, especially when you can avoid them in a small cost. What client app can do is to handle the error, terminate the tread gracefully (save the state, notify users, write debug logs, send debug info to developer) and restart the thread.

src/stream.rs
dr-orlovsky: Agree, will update

src/raw_client.rs
dr-orlovsky: Agree, will update

src/types.rs
afilini: Yeah I was suggesting to add a specific variant that reflects the true meaning of the error, and then the user can choose how to handle that variant specifically.

I generally agree with you that it's better to return an error when possible, but on the other end it's very important to ensure that the code can't accidentally end up in an inconsistent/unexpected state: in that regard, the panics on `lock()` were used to simplify things a little bit.

Basically we chose to panic rather than potentially incur in some kind of logic bug, which I still think it's a better trade-off in many cases.

dr-orlovsky: I am ok with adding error-specific variant. Or you are against removing `unwrap()`s on `lock()`?

dr-orlovsky:
If possible will prefer adding error-specific variants after #36 merge, so I can rebase this PR on it and add error displays at the same time

src/types.rs
afilini: No that's fine, I'm just saying that now that we don't panic we have to be a bit more careful and think about the potential edge cases..

dr-orlovsky:
@afilini rebased & fixed according to your comments

afilini (approved):

PR #34, opened 2021-01-21
Streamlining batch_call and removing unnecessary unwraps()

dr-orlovsky:
This improves overall function code style & allows to remove two unnecessary unwraps and one implicit panic. Together with #33 this makes this library unwrap-free and stable (i.e. not causing the software using it to unexpectedly terminate instead or reporting error to the calling function)

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -683,10 +683,7 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
 
         for req_id in missing_responses.iter() {
             match self.recv(&receiver, *req_id) {
-                Ok(mut resp) => answers.insert(
-                    resp["id"].as_u64().unwrap_or_default(),
-                    resp["result"].take(),
-                ),
+                Ok(mut resp) => answers.insert(req_id, resp["result"].take()),
                 Err(e) => {
                     // In case of error our sender could still be left in the map, depending on where
                     // the error happened. Just in case, try to remove it here
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -2,7 +2,7 @@
 //!
 //! This module contains the definition of the raw client that wraps the transport method
 
-use std::collections::{HashMap, HashSet, VecDeque};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque};
 use std::io::{BufRead, BufReader, Read, Write};
 use std::mem::drop;
 use std::net::{TcpStream, ToSocketAddrs};
@@ -647,8 +647,8 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
     fn batch_call(&self, batch: &Batch) -> Result<Vec<serde_json::Value>, Error> {
         let mut raw = Vec::new();
 
-        let mut missing_responses = HashSet::new();
-        let mut answer = Vec::new();
+        let mut missing_responses = BTreeSet::new();
+        let mut answers = BTreeMap::new();
 
         // Add our listener to the map before we send the request, Here we will clone the sender
         // for every request id, so that we only have to monitor one receiver.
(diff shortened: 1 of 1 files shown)


src/raw_client.rs
```diff
@@ -668,32 +667,32 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
                     warn!("got error for req_id {}: {:?}", req_id, e);
                     warn!("removing all waiting req of this batch");
-                    let mut guard = self.waiting_map.lock().unwrap();
-                    for req_id in missing_responses.iter() {
-                        guard.remove(&req_id);
-                    }
+                    let mut guard = self
+                        .waiting_map
+                        .lock()
+                        .map_err(|_| Error::CouldntLockReader)?;
```
afilini, line 681: Same as in #33, this needs a specific error

src/raw_client.rs
```diff
@@ -668,32 +667,32 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
+                    guard.remove(&req_id);
```
afilini, line 682: I think I'm missing something, why did you remove the loop here?

src/raw_client.rs
```diff
@@ -668,32 +667,32 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
                 }
-            };
-            let resp_id = resp["id"].as_u64().unwrap() as usize;
-
-            missing_responses.remove(&resp_id);
-            answer.push(resp);
+                Ok(resp) => {
+                    let mut res = resp
+                        .get("result")
+                        .cloned()
```
afilini, line 688: I would keep this part as it was before, just `take()`-ing the result and adding it to `answer`.

There's no need to return an `InvalidResponse` here because if there's something wrong with it we'll just see that later when trying to parse the response into a specific struct.

I guess if you really wanted to return that error instead of a generic `serde` error you could add this to `impl_batch_call`:

 https://github.com/bitcoindevkit/rust-electrum-client/blob/af10aeac99bcd2b68c1b84b264d0971c76873301/src/raw_client.rs#L49

You could map the error here to `InvalidResponse` before returning it

src/raw_client.rs
```diff
@@ -668,32 +667,32 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
+                    answers.insert(req_id, res.take());
+                }
+            }
         }
 
-        answer.sort_by(|a, b| a["id"].as_u64().partial_cmp(&b["id"].as_u64()).unwrap());
-        let answer = answer.into_iter().map(|mut x| x["result"].take()).collect();
-
-        Ok(answer)
+        Ok(answers.into_iter().map(|(_, item)| item).collect())
```
afilini, line 695: I think you are missing the part where you remove the `id` of the response we've just received from `missing_responses`.

afilini (changes requested):

src/raw_client.rs
dr-orlovsky: Yep,my bad, I haven't understood the code and thought `req_id` is an external variable. Will fix. Now I see why you was maintaining list of missing responses up to date.

In general, can you explain what actually should happen here? What should be a consequences of erroring one of the batch request? Cancellation of the rest of the requests and returning single error, ignoring all ok-requests already received? That does not make a lot of sense for me... May be I am missing something?

afilini: Right now if one request fails it makes the whole batch fail. If there are still some requests pending it will just ignore their response when they come back (there's no way of cancelling a request you've already sent).

This line specifically is where all the pending requests id are removed from the map of responses we are waiting for, which means that once the response comes in the current `reader_thread` will just ignore it:

https://github.com/bitcoindevkit/rust-electrum-client/blob/02462b37b383ad9787ad95f53bcf8b5e5d17f616/src/raw_client.rs#L493-L500

dr-orlovsky:
Fixed according to suggestions & improved

src/raw_client.rs
```diff
@@ -681,32 +681,27 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
         self.increment_calls();
 
-        while !missing_responses.is_empty() {
-            let req_id = *missing_responses.iter().next().unwrap();
-            let resp = match self.recv(&receiver, req_id) {
-                Ok(resp) => resp,
+        for req_id in missing_responses.iter() {
+            match self.recv(&receiver, *req_id) {
+                Ok(mut resp) => answers.insert(
+                    resp["id"].as_u64().unwrap_or_default(),
```
afilini, line 687: I'm not entirely sure `unwrap_or_default()` is a good idea here: in theory if we got to this point it means that the response had an `id`, but in case it's missing assuming its value is `0` can create issues. I think it would be better to replace it with `.unwrap_or(req_id)`, or even just always returning the `req_id` directly.

dr-orlovsky: You are right. Fixed.

afilini (approved):

PR #37, opened 2021-01-21
Improving error type

dr-orlovsky:


src/types.rs
```diff
@@ -264,6 +265,7 @@ pub struct ScriptNotification {
 
 /// Errors
 #[derive(Debug)]
+#[non_exhaustive]
```
afilini, line 268: Isn't it better to leave this out? I'm thinking about somebody trying to catch some specific errors: if the enum is `non-exhaustive` then they'll have a "wildcard' branch for types that could be added in the future, and they might not realize that a new error they are interested in has been added.

Without this, they can match every single error if they want to, and if a new one is added they will get a compilation error informing them about it

dr-orlovsky: There is another side of this story, which we witnessed in rust-bitcoin: not doing errors non-exhaustive leads to the fact that addition of each new error type is a breaking API change and require major version release, preventing users from upgrading and leaving behind new features from being used. So for early-stage software before error API is stabilized and new features added quickly sometime desirable to have non-exhaustive errors; otherwise all error and code improvements will be constantly postponed "because they are breaking changes".

afilini: That's a good point, but I'd still rather leave it "exhaustive". I prefer to force users to manually handle errors compared to potentially adding some new variants silently.

dr-orlovsky, line 268:
```suggestion
```

dr-orlovsky: Done. Actually I used in-place removal of the line with GitHub suggestion & commit new function for the first time :)

afilini: Yeah that's not bad, but the downside is that they sign your commit with their key. If that's okay for you I can merge it like this, otherwise you should sign your commit locally and force-push

dr-orlovsky: I think with such tiny change that has no importance it's ok to use their signing key :)

PR #32, opened 2021-01-20
Fixing calls_made fn signature

dr-orlovsky:
Currently compilation with `--all-features` and `--feature debug-calls` or `cargo test` is broken. This PR fixes it

PR #38, opened 2021-01-21
Fix CI testing for all features

dr-orlovsky:
`cargo test --all` is a deprecated equivalent of `cargo test --workspace`. This leads to the fact that tests are run without using all features (utilizing default set of features only) and such bugs as #32 missed by CI.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Fmt
         run: cargo fmt -- --check --verbose
       - name: Test
-        run: cargo test --verbose --all
+        run: cargo test --verbose --all-features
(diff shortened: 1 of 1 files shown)


dr-orlovsky:
CI is failing due to #32. This PR is intended for detecting the problem missed before, so this is ok to fail.

Now. with #32 being merged if you re-run CI you will get success

dr-orlovsky:
I think this is the most important across all the PRs in terms of moving forward with the rest: w/o it we can't check whether new PRs does not break compilation (since currently CI does not cover all of the code)

afilini:
I think this needs to be rebased first to make the CI pass

dr-orlovsky:
Rebased

PR #48, opened 2021-02-16
Rework retry logic within `impl_inner_call`

### LINKED ISSUE #47: Off-by-one error for retries?
The default value for `retries` in the `Config` is 1.

The retry value is used here:

https://github.com/bitcoindevkit/rust-electrum-client/blob/7db9a1331b7bacce7c21b51582c3ee006a850985/src/client.rs#L57-L59

I think there is an off-by-one error here because effectively, a value of 1 means 0 **re**-tries because we fail after the first error so we never actually **re**-try the call. If we were to use `>` instead of `==` then we would actually retry the RPC call.

thomaseizinger:
The first patch fixes #47 by properly comparing the number of failed attempts with the value of config.retry.

The second patch improves the logging that happens as part of this logic.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -77,11 +77,16 @@ macro_rules! impl_inner_call {
                                     break;
                                 },
                                 Err(e) => {
-                                    warn!("client retry:{}/{} {:?}", errors.len() + 1, $self.config.retry(), e);
-                                    errors.push(e);
-                                    if errors.len() as u8 == $self.config.retry() {
+                                    let failed_attempts = errors.len() + 1;
+
+                                    if retries_exhausted(failed_attempts, $self.config.retry()) {
+                                        warn!("re-creating client failed after {} attempts", failed_attempts);
                                         return Err(Error::AllAttemptsErrored(errors));
                                     }
+
+                                    warn!("re-creating client failed with {}, retry: {}/{}", e, failed_attempts, $self.config.retry());
+
+                                    errors.push(e);
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -52,12 +52,17 @@ macro_rules! impl_inner_call {
                     return res;
                 },
                 Err(e) => {
-                    warn!("call retry:{}/{} {:?}", errors.len() + 1 , $self.config.retry(), e);
-                    errors.push(e);
-                    if errors.len() as u8 == $self.config.retry() {
+                    let failed_attempts = (errors.len() + 1) as u8;
+
+                    if failed_attempts > $self.config.retry() {
```
RCasatta, line 57: I think in the rather subtle case of `retry = 255` the if condition will never satisfy (or panic in debug mode)

It may be better to let `failed_attempts` usize and cast retry instead

thomaseizinger: Using `usize` for the `retry` value feels wrong, the value of `retry` is independent from the platforms pointer width. I'll think of something.

thomaseizinger: See 3d08748.

thomaseizinger:
I saw that there is also some retry logic for _creating_ a new client. Should we adapt that too?

RCasatta:
ack cb9b51f3562e87573b91fb9f39f69f0227e45a2d

Yes, I think we need to adapt the client creation too

we can merge this and open another PR for that or add to this, whatever comes first

thomaseizinger:
> Yes, I think we need to adapt the client creation too

See c8bace9.

Like before, this is reusing the `errors` vector to track the errors of re-creating the client as well. I am not fully convinced that this is the best way or whether we should actually isolate that.

RCasatta:
ack c8bace9407ba97fff63d4d474469014207cb82c3

PR #50, opened 2021-02-22
Implement timeouts when there are multiple socket addrs

afilini:


RCasatta:
very nice!

I wonder if we can write a test instantiating a raw client with two endpoints, the first one being a locally started socket emulating a timeout

afilini:
I don't know what's the best way to simulate a timeout, I'll look into that. Here we'd like to test the connection timeout, not a message timeout, so just opening a socket and connecting to it doesn't work.

We need something that silently drops packets instead of replying with an error like connection refused.

PR #55, opened 2021-07-30
Enable TCP support without socks5 or SSL

tcharding:
Currently users of the library must configure SSL or a socks5 proxy even if they only want a plain TCP stream. This is contrary to the comment in `lib.rs` that suggests `features=minimal` may be used to get TCP only (will not currently build with `--no-default-features`).

There is no need to force plaintext TCP stream users to enable SSL and/or proxy features.

Enable `Client` to be used for plaintext TCP stream without requiring any of the features `use-openssl`, `use-rustls`, or `proxy`.

- Patch 1 clears a few clippy warnings by adding trivial docs (I'm not 100% sure the link (`[Config]`) works.
- Patch 2 implements the actual PR as describe above.

Note: All the conditional compilation is pretty ugly IMO. I use additional braces that are not strictly required by rustc but seem to make the code easier to read (also my editor messes with the syntax highlighting without them, bad form I know doing code changes to assist the editor).

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -8,7 +8,7 @@ use bitcoin::{Script, Txid};
 
 use api::ElectrumApi;
 use batch::Batch;
-use config::Config;
+use config::{Config, ConfigBuilder};
 use raw_client::*;
 use std::convert::TryFrom;
 use types::*;
@@ -16,14 +16,14 @@ use types::*;
 /// Generalized Electrum client that supports multiple backends. This wraps
 /// [`RawClient`](client/struct.RawClient.html) and provides a more user-friendly
 /// constructor that can choose the right backend based on the url prefix.
-///
-/// **This is available only with the `default` features, or if `proxy` and one ssl implementation are enabled**
 pub enum ClientType {
-    #[allow(missing_docs)]
+    /// TCP stream client.
     TCP(RawClient<ElectrumPlaintextStream>),
-    #[allow(missing_docs)]
+    /// SSL stream client.
+    #[cfg(any(feature = "openssl", feature = "rustls"))]
     SSL(RawClient<ElectrumSslStream>),
-    #[allow(missing_docs)]
+    /// Socks5 proxy client (either SSL or TCP).
+    #[cfg(feature = "proxy")]
     Socks5(RawClient<ElectrumProxyStream>),
 }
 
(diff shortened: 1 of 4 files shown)


afilini:
In principle I think it makes sense to support just the TCP client (although it can be done already by using the `RawClient` directly), but I don't really like having public-facing structs with fields that are added conditionally.

I think always having them would also clean up some of the ugly conditional compilation stuff, because instead of conditionally adding blocks of code we can just use `if cfg!() {}`.

I guess the only argument for having them conditionally added is that more mistakes could be caught at compile time, but since the `Client` structure is mainly configured through a string there are always gonna be runtime errors, so maybe it's not worth the extra complexity.

tcharding:
> In principle I think it makes sense to support just the TCP client (although it can be done already by using the `RawClient` directly), but I don't really like having public-facing structs with fields that are added conditionally.

Face palm, I did not think of this. If this is the case then this PR is totally pointless.

> I think always having them would also clean up some of the ugly conditional compilation stuff, because instead of conditionally adding blocks of code we can just use `if cfg!() {}`.

I did try playing with `cfg!` but couldn't get it to work.

> I guess the only argument for having them conditionally added is that more mistakes could be caught at compile time, but since the `Client` structure is mainly configured through a string there are always gonna be runtime errors, so maybe it's not worth the extra complexity.

I'm going to close this for now and just use RawClient. Thanks for the review, and sorry for the noise.

afilini:
No worries! :)

I think it's always good to try new things, in this case I think we've also tried doing something similar right at the beginning but we couldn't really make it work, so we opted for a low-level but very flexible interface plus a more high-level interface for the general public.

tcharding:
> In principle I think it makes sense to support just the TCP client (although it can be done already by using the RawClient directly), but I don't really like having public-facing structs with fields that are added conditionally.

It was bothering me why I'd not realized this and just now I worked it out. Just sharing my motivations for the changes incase they seem strange :)

I'm trying to build/maintain a stack of rust bitcoin libs with minimal patch sets that add altcoin support. In doing so I wanted to use the `electrum` module in `bdk` with as few changes as possible and this desire led me to hacking the `client` module here in `electrum-client`. I'm going to have a go at instead hacking `dbk` now you've pointed this out!

Thanks again for your patient reviews.

PR #58, opened 2021-08-22
Support getting verbose transactions

TheCharlatan:
This allows the caller to get extra information of a transaction, like the block hash of the block its included in and the number of
confirmations.

Some electrum servers do not support this method, which is reflected in the test of the method.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -33,6 +33,9 @@ pub trait ElectrumApi {
         Ok(deserialize(&self.transaction_get_raw(txid)?)?)
     }
 
+    /// Gets the verbose transaction with `txid`. Returns an error if not found.
+    fn transaction_get_verbose(&self, txid: &Txid) -> Result<GetTransactionVerboseRes, Error>;
+
     /// Batch version of [`transaction_get`](#method.transaction_get).
     ///
     /// Takes a list of `txids` and returns a list of transactions.
--- a/src/client.rs
+++ b/src/client.rs
@@ -253,6 +253,11 @@ impl ElectrumApi for Client {
     }
 
     #[inline]
+    fn transaction_get_verbose(&self, txid: &Txid) -> Result<GetTransactionVerboseRes, Error> {
+        impl_inner_call!(self, transaction_get_verbose, txid)
+    }
+
+    #[inline]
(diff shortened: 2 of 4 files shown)


johncantrell97:
Anything holding this up? I would find this useful.

tnull:
+1 for this. On the LDK side of things we might soon need this and/or find a workaround.

notmandatory:
I just put this on the BDK roadmap so we can get some attention on it. Any others like this one you want to get some attention on let me know.

afilini:
We recently merged #74 so that users can make raw calls without having to wait for us to implement them.

Now that we have that I'm not sure if it makes sense to merge methods that aren't supported by the all the main implementations, I think I would rather let people who know what they are doing implement this themselves in their own code.

tnull:
> We recently merged #74 so that users can make raw calls without having to wait for us to implement them.
> 
> Now that we have that I'm not sure if it makes sense to merge methods that aren't supported by the all the main implementations, I think I would rather let people who know what they are doing implement this themselves in their own code.

Hm, but `transaction_get_verbose` is just a regular call that an `ElectrumClient` should support? If you're talking about the BDK side of things I understand it's debatable what should or should not be part of the `Blockchain` interface, but IIUC there is currently no way to access the underlying client once once it's passed to `Blockchain::from_client`? So the users can't then implement their own logic on top of it?

afilini:
Yeah my point was that apparently not all electrum implementations support this call, so maybe adding it to the client is a bit misleading because people would expect it to work everywhere. Maybe they test with just an implementation and it looks fine and then it breaks somewhere else. Since this is not an "ElectrumX" or "electrs" client specifically, we try to only implement stuff that work everywhere.

> there is currently no way to access the underlying client once once it's passed to Blockchain::from_client

I think `ElectrumClient` should implement `Deref<Target=Client>`, if it doesn't I'll open a quick PR now

TheCharlatan:
> We recently merged https://github.com/bitcoindevkit/rust-electrum-client/pull/74 so that users can make raw calls without having to wait for us to implement them.
> Now that we have that I'm not sure if it makes sense to merge methods that aren't supported by the all the main implementations, I think I would rather let people who know what they are doing implement this themselves in their own code.

This is reasonable, I'm closing this PR again.

PR #61, opened 2022-01-11
Test timeout

RCasatta:
I noticed timeout parameter is not approximately equal to the time measured with the included test.

timeout=1 -> elapsed 4s
timeout=2 -> elapsed 6s
timeout=3 -> elapsed 8s
timeout=4-> elapsed 10s

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -392,4 +392,37 @@ mod tests {
         assert!(client.is_ok());
         assert!(elapsed > Duration::from_secs(2));
     }
+
+    #[test]
+    fn test_timeout() {
+        use std::thread;
+        use std::time::{Duration, Instant};
+
+        let listener = std::net::TcpListener::bind(("127.0.0.1", 0)).unwrap(); // 0 means the OS choose a free port
+        let endpoint = format!("{}", listener.local_addr().unwrap());
+
+        let (s, r) = std::sync::mpsc::channel();
+        thread::spawn(move || {
+            // emulate an electrum server socket not replying for 30 seconds
+            s.send(());
+            let (mut s, _) = listener.accept().unwrap();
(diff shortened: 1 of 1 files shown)


afilini:
Ok I spent some time debugging this: I was wrong when I told you that you need to send an explicit ping, it look like this is done internally right after connecting to the server.

So this explains why we have the "2x timeout" part of the final time.

What I still don't understand is why there's the "+ 2" part. Looking at strace I can see that there's a 2s sleep somewhere right after it fails to receive anything (you can see `recvfrom` returning after 1s with `EAGAIN`)

```
[pid 16177]      0.000134 sendto(5, "{\"jsonrpc\":\"2.0\",\"id\":0,\"method\":\"server.ping\",\"params\":[]}\n", 60, MSG_NOSIGNAL, NULL, 0) = 60
[pid 16177]      0.000197 recvfrom(5, 0x77d448000fc0, 8192, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)
[pid 16177]      1.047175 clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=2, tv_nsec=0}, 0x77d44faed480) = 0 (????)
[pid 16177]      2.000259 socket(AF_INET, SOCK_STREAM|SOCK_CLOEXEC, IPPROTO_IP) = 6
```

afilini:
Ah, got it, that's the retry timeout:

https://github.com/bitcoindevkit/rust-electrum-client/blob/98e1bf205c609a75b0261c390f88e1a4cd8e390a/src/client.rs#L72

Since we push the error before sleeping, it will sleep for `1 << 1` so two seconds

notmandatory:
This is pretty old, is it still something we need and should I target this for the next release?

PR #62, opened 2022-02-04
Upgrade rustls and webpki version

thomaseizinger:


src/raw_client.rs
```diff
@@ -277,19 +280,23 @@ impl RawClient<ElectrumSslStream> {
...
-            _dns_name: webpki::DNSNameRef<'_>,
-            _ocsp: &[u8],
-        ) -> Result<rustls::ServerCertVerified, rustls::TLSError> {
-            Ok(rustls::ServerCertVerified::assertion())
+            end_entity: &Certificate,
+            intermediates: &[Certificate],
+            server_name: &ServerName,
+            scts: &mut dyn Iterator<Item = &[u8]>,
+            ocsp_response: &[u8],
+            now: SystemTime,
```
afilini, line 297: You should prefix these arguments with the underscore since they are unused

thomaseizinger: Can do!

If we are about this then I think CI should catch such stuff :)

afilini: I later noticed that there are a few other warnings, so yes we should consider checking for them in the CI. In this case I pointed it out because we were using the underscores before and I saw that you removed them.

afilini (approved):
ACK 68d2f233064c5f4d37f0c717ad9d8f56af5cf4fa

PR #71, opened 2022-06-24
Add `Config::builder` method

evanlinjin:
The intention is to reduce some confusion on how to manipulate `Config`.
The alternative is to make `Config` fields public.


afilini (approved):
ACK aea9da234bf3ac4ca026833ebbc0ffb122bb06f9

PR #74, opened 2022-07-27
Add raw_call to ElectrumApi

zoedberg:
In order to give more flexibility to the users of this library I added a `raw_call` method to the `ElectrumApi` trait that will allow to call any arbitrary electrum API call.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -65,6 +65,9 @@ pub trait ElectrumApi {
         self.transaction_broadcast_raw(&buffer)
     }
 
+    /// Executes the requested API call returning the raw answer.
+    fn raw_call(&self, call: &Call) -> Result<serde_json::Value, Error>;
+
     /// Execute a queue of calls stored in a [`Batch`](../batch/struct.Batch.html) struct. Returns
     /// `Ok()` **only if** all of the calls are successful. The order of the JSON `Value`s returned
     /// reflects the order in which the calls were made on the `Batch` struct.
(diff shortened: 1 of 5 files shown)


afilini (approved):
ACK 3d96644a0f7e5823b8b9bd712640ddfbb253061d

dr-orlovsky:
Am I correct that after the latest version was yanged this is not yet released in any working crate version?

src/api.rs
```diff
@@ -65,6 +65,9 @@ pub trait ElectrumApi {
         self.transaction_broadcast_raw(&buffer)
     }
 
+    /// Executes the requested API call returning the raw answer.
+    fn raw_call(&self, call: &Call) -> Result<serde_json::Value, Error>;
```
dr-orlovsky, line 69: this leads to calling method like `raw_call(&("method_name".to_owned(), vec![]))`, which is very non-rust ideomatic. Thus I propose a follow-up PR with changing function signature to `raw_call(&self, method_name: &str, params: impl IntoIterator<Item=Param>)`.

This, however, will be a breaking change and will require a new minor version bump.

dr-orlovsky (commented):
Post-merge review with improvement proposal

src/api.rs
afilini: Opened https://github.com/bitcoindevkit/rust-electrum-client/pull/84 to implement your proposal

PR #76, opened 2022-08-02
Implement timeout for socks connection

RCasatta:
To achieve this we have to include a modified socks library, with supports for
timeout

Remove the error `BothSocksAndTimeout` which isn't needed anymore

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,16 +23,23 @@ serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 
 # Optional dependencies
-socks = { version = "0.3", optional = true }
 openssl = { version = "0.10", optional = true }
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 
 # Optional dependencies
-socks = { version = "0.3", optional = true }
```
afilini, line 26: I don't think I can publish to crates.io with a local dependency. I think you need to move `socks` inside our `src` folder and include it as a local module

RCasatta: ahah ooops

RCasatta: `socks` included in `src`

RCasatta:
not necessary anymore

RCasatta:
I closed the wrong one

Issue #67, opened 2022-05-27
Problems when running multiple batch-calls in parallel


da-kami:
This is is a bit tricky to describe.

Recently we see quite a lot of these warnings in our logs:

```
2022-05-27 03:32:23  WARN electrum_client::raw_client: got error for req_id 81: JSON(Error("EOF while parsing a value", line: 1, column: 0))
2022-05-27 03:32:23  WARN electrum_client::raw_client: removing all waiting req of this batch
2022-05-27 03:32:23  WARN electrum_client::client: call 'batch_script_get_history' failed with EOF while parsing a value at line 1 column 0, retry: 1/1
```

At first we did not pay much attention to this, but it actually is quite a problem, because the batch processing fails at this point and keeps failing. We just reconnect and keep processing until we run into an error again.

I traced the problem back to this call:

https://github.com/bitcoindevkit/rust-electrum-client/blob/master/src/raw_client.rs#L491

in the `raw_client` code, context:

https://github.com/bitcoindevkit/rust-electrum-client/blob/master/src/raw_client.rs#L479-L495

I found a pattern when this problem happens. It is related to triggering multiple batch-processing calls in parallel.
Whenever a new "group" of a batch is handled as part of one request the next request will receive an empty response from the server. Since the response is literally just `""` (an empty string) the conversion into `serde::Value` will fail.

Here is some analysis from the `electrum_client` `trace` logs:

```
2022-05-27 03:31:23 TRACE electrum_client::raw_client: Reader thread 77 received a response for its request
2022-05-27 03:31:23 TRACE electrum_client::raw_client: Starting reader thread with `until_message` = Some(78)
2022-05-27 03:31:23 TRACE electrum_client::raw_client: <== {"id":78,"jsonrpc":"2.0","result":[]}

2022-05-27 03:31:23 TRACE electrum_client::raw_client: Reader thread 78 received a response for its request
2022-05-27 03:31:23 TRACE electrum_client::raw_client: Starting reader thread with `until_message` = Some(79)
2022-05-27 03:31:23 TRACE electrum_client::raw_client: <== {"id":79,"jsonrpc":"2.0","result":[]}

2022-05-27 03:31:23 TRACE electrum_client::raw_client: Reader thread 79 received a response for its request
2022-05-27 03:31:23 TRACE electrum_client::raw_client: ==> {"jsonrpc":"2.0","id":80,"method":"blockchain.headers.subscribe","params":[]}
[...]

PR #78, opened 2022-08-03
Handle EOF in the reader thread

### LINKED ISSUE #70: Wallet sync can get stuck when ignoring empty lines
Hi,
after running a fork of this project (older version, due to older BDK) with one of the changes changes backported (https://github.com/bitcoindevkit/rust-electrum-client/commit/0e3e6325e65749c0446fbede99d763b98321362d), we have noticed some issues with the wallet sync hanging:
https://github.com/itchysats/itchysats/issues/2287

Unfortunately, we couldn't reproduce the error locally to provide more info or contribute a fix :( 
We ended reverting the fix locally, and the error is not happening in anymore. 

Our fork diff with the recent master:
https://github.com/da-kami/rust-electrum-client/compare/fix/ignore-empty-lines...bitcoindevkit:master

We aim to bump bdk stack soon, we'll see whether the issue will happen again with `0.10.1`.

afilini:
Instead of ignoring empty lines we should return an error, because it means we are at EOF.

Fixes #70 and hopefully also #67

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -477,20 +477,22 @@ impl<S: Read + Write> RawClient<S> {
 
                 // Loop over every message
                 loop {
-                    raw_resp.clear();
-
-                    if let Err(e) = reader.read_line(&mut raw_resp) {
+                    let send_err = |e: std::io::Error| {
                         let error = Arc::new(e);
                         for (_, s) in self.waiting_map.lock().unwrap().drain() {
                             s.send(ChannelMessage::Error(error.clone()))?;
                         }
-                        return Err(Error::SharedIOError(error));
-                    }
-                    trace!("<== {:?}", raw_resp);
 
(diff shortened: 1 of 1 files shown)


RCasatta:
This test https://github.com/Blockstream/gdk/blob/025299de1c9a593c66c7f27140fe1d6816aa732a/subprojects/gdk_rust/gdk_rust/tests/integration.rs#L1658 testing timeout with a non-replying server hangs without this PR, while it works with this PR.
ACK acc05c3666421f0a03c140b39d10baa924e89d44

afilini:
I had to yank the release that contained this commit because it broke our ci. To be honest I don't know what to do about this anymore, our original code used to work fine. I never managed to reproduce the empty lines issue myself, I'm considering just moving back to the old thing and being done with this

PR #80, opened 2022-08-16
Client: Drop script subscriptions to free lock

TheCharlatan:
This fixes a supposed deadlock when the reader thread receives a script subscription notification from the server before it receives a subscribing response. 

I am not 100% sure if this actually fixes a deadlock, but I ran into this in a regtest setup where a bunch of subscriptions were made, while some script notifications could be processed in parallel. My test failed sporadically with the electrum client locking up completely and adding this drop made the tests consistently pass again.


afilini (approved):
ACK aa0b9578188dcae8219316ef392329d764b04b2e, makes a lot of sense!

PR #88, opened 2022-09-05
`timeout()` and `socks()` don't return Result

RCasatta:
The `Result` was necessary in the past because they couldn't both be set,
this is not true anymore after 7493630ed88af5558b84a1f715890f67cfa4ec24


afilini:
Needs a cargo fmt and then I will ACK this!

RCasatta:
force pushed with cargo fmt

afilini (approved):
ACK 561db261d3529a058c7c94a8f48852c637b437c7

PR #93, opened 2023-01-05
Adding optional tests …

ulrichard:
… to showcase that it currently only works with electrs servers


notmandatory:
Hi, i'm just getting caught up on PRs in this repo. It's good to have a way to show which electrum servers this repo works and doesn't work with, but I'd rather not have a test that always fails. Also I'd rather not use non-local servers that could fail for other reasons.  How would you feel about getting this to at least pass with `electrsd` and commenting out the other servers as work for future PRs when and if we can get those other servers working and locally testable?

PR #94, opened 2023-02-14
Add `Batch::raw` and improve docs

evanlinjin:
Being able to add raw requests to `Batch` makes sense from an API standpoint because we already allow raw non-batched requests. This is also useful when the electrum server API gets an updated version and our client is unable to keep up.

Additional to this, I have improved the documentation and made `Call` private (since `Call` is never used externally).

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -22,6 +22,11 @@ pub struct Batch {
 }
 
 impl Batch {
+    /// Add a raw request to the batch queue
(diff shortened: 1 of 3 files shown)


evanlinjin:
Maybe it would be better to make `Batch::calls` public. This will allow us to `collect` into it, which would be more useful?

notmandatory (approved):
ACK adb0c7444e610570ced1660c1890a5c1fdbd48c8

Issue #98, opened 2023-02-28
`ScriptStatus` should be also `Serializable`


RCasatta:
Implementors should persist the `ScriptStatus` to compare for changes between runs

RCasatta:
Not necessarily, I now think it's better to recompute the status from the transactions in the cache

RCasatta:
Thinking again, it makes sense to `Serialize` `ScriptStatus` otherwise the cache should also store the order of received transactions (you can order txs if they are at different height, but if they are in the same block you can't in edge cases...) but it's more convenient to simply cache the `ScriptStatus`

PR #101, opened 2023-04-05
Make ScriptStatus Serializable

RCasatta:
it's common to cache this value client side, so making it Serializable simplify things downstream


RCasatta:
close https://github.com/bitcoindevkit/rust-electrum-client/issues/98

afilini (approved):
ACK 80b2adeb4b314261d941f062d310a495c3fe44d2

PR #106, opened 2023-05-12
Retry client should return AlreadySubscribed error

RCasatta:
Users may leverage the Error to avoid remembering client-side the subscription status. eg. Always subscribing and calling script_pop in case AlreadySubscribed is returned.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -49,7 +49,7 @@ macro_rules! impl_inner_call {
             drop(read_client);
             match res {
                 Ok(val) => return Ok(val),
-                Err(Error::Protocol(_)) => {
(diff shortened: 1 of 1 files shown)


LeoComandini:
ACK 26738931b9fb6eb484e3d4a7bb9f8a3ade8d6938, code review and tested locally.

This makes the comment [here](https://docs.rs/electrum-client/latest/electrum_client/trait.ElectrumApi.html#tymethod.script_subscribe) _"Returns [Error::AlreadySubscribed](https://docs.rs/electrum-client/latest/types/enum.Error.html#variant.AlreadySubscribed) if already subscribed to the script."_ correct.

danielabrozzoni:
Closing as duplicated of #107

PR #107, opened 2023-05-12
Retry client should return AlreadySubscribed error

RCasatta:
Users may leverage the Error to avoid remembering client-side the subscription status. eg. Always subscribing and calling script_pop in case AlreadySubscribed is returned.

Backport for version 0.14.0

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -49,7 +49,7 @@ macro_rules! impl_inner_call {
             drop(read_client);
             match res {
                 Ok(val) => return Ok(val),
-                Err(Error::Protocol(_)) => {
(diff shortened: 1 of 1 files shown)


afilini:
This is technically a breaking change because the same error returned to 0.15 (or 0.14 once we backport) would lead to a different outcome for the user. That said, I consider this more of a bug fix than a simple improvement: it doesn't make sense to keep retrying when we get the "already subscribed" error, it's definitely not a transient error that may get fixed with the next try.

So I think it makes sense to merge this and release as 0.15.1 and then backport and release 0.14.1

afilini:
ACK 3a61037733f700c0b03de4069ba055ffbd74254f

PR #110, opened 2023-06-27
Update rustls dependency

tcharding:
Update the `rustls` dependency to version 0.21 and bump the version of this crate to 0.16 so the change can be released.

I don't know what stage in your release cycle you guys are up to so I put the version bump as a separate patch - can drop it if not needed.

Done to help with https://github.com/bitcoindevkit/rust-esplora-client/pull/51


danielabrozzoni (approved):
utACK f5a438cd0a5b68b3f74d3c144e90e2b099c21261

PR #122, opened 2023-10-27
Add utility to validate `GetMerkleRes`

tnull:
I recently needed to validate a Merkle inclusion proof as retrieved via `transaction_get_merkle`. 

As I figured it might be useful to other people, too, we add it here as a simple utility method.


notmandatory (approved):
ACK 54fd52d898bd8cfd448a4d303f12c156a7a33201

Thanks for the addition, it's a handy thing to have in this project.

PR #128, opened 2023-12-29
ci: add test with MSRV 1.63.0

notmandatory:
Since the main BDK crates are changing to MSRV 1.63.0 as is LDK the CI for this project should also have an MSRV and test against it in CI.

Also rustc 1.63 is the version shipped with the current debian stable (bookworm): https://packages.debian.org/stable/rust/rustc

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -3,13 +3,17 @@ on: [push, pull_request]
 name: CI
 
 jobs:
-
   test-fmt:
     name: Test
     runs-on: ubuntu-20.04
     env:
       TEST_ELECTRUM_SERVER: electrum.blockstream.info:50001
       #TEST_ELECTRUM_SERVER: bitcoin.aranguren.org:50001
+    strategy:
+      matrix:
+        rust:
+          - stable # STABLE
+          - 1.63.0 # MSRV
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


oleonardolima:
@notmandatory I did another pass on this one, I think this would also be a good one to get into the next release, to keep the project in standard with the rest of the ecosystem, wdyt ?

oleonardolima (approved):
utACK 4b421084e912cb7395ea49c1bc7955b445270feb

.github/workflows/cont_integration.yml
```diff
@@ -24,7 +28,7 @@ jobs:
       - name: Install rustup
         run: curl https://sh.rustup.rs -sSf | sh -s -- -y
       - name: Set default toolchain
```
storopoli, line 30: Maybe change this to "MSRV toolchain"?

oleonardolima: hmm, but it also gonna use the stable one due to the matrix strategy 🤔

notmandatory: This sets the "default" toolchain (stable or 1.63.0) for each test run so I think the name is fine.

storopoli (approved):
ACK 4b42108

ValuedMammal (approved):
ACK 4b421084e912cb7395ea49c1bc7955b445270feb looks good to me

PR #132, opened 2024-04-17
Upgrade rustls to 0.23

nickfarrow:
With rustls 0.23 there is no longer a dependency on ring, allowing easier compilation for various targets.

Not super confident with my updates to `ServerCertVerifier` and `Der` of certificates (is this being tested?), needs review.

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -21,12 +21,15 @@ use bitcoin::{Script, Txid};
 
 #[cfg(feature = "use-openssl")]
 use openssl::ssl::{SslConnector, SslMethod, SslStream, SslVerifyMode};
+
 #[cfg(all(
     any(feature = "default", feature = "use-rustls"),
     not(feature = "use-openssl")
 ))]
 use rustls::{
-    ClientConfig, ClientConnection, OwnedTrustAnchor, RootCertStore, ServerName, StreamOwned,
+    pki_types::ServerName,
+    pki_types::{Der, TrustAnchor},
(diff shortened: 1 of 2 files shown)


RCasatta:
Version `0.21` has also other issues...

```
Crate:     rustls
Version:   0.21.7
Title:     `rustls::ConnectionCommon::complete_io` could fall into an infinite loop based on network input
Date:      2024-04-19
ID:        RUSTSEC-2024-0336
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0336
Severity:  7.5 (high)
Solution:  Upgrade to >=0.23.5 OR >=0.22.4, <0.23.0 OR >=0.21.11, <0.22.0

```

src/raw_client.rs
```diff
@@ -21,13 +21,12 @@ use bitcoin::{Script, Txid};
 
 #[cfg(feature = "use-openssl")]
 use openssl::ssl::{SslConnector, SslMethod, SslStream, SslVerifyMode};
+use rustls::pki_types::{Der, TrustAnchor};
```
RCasatta, line 24: As seen from failing CI jobs, this must be gated

notmandatory (approved):
utACK 28b1aaa0c304c52eb98d13ddc4bcc9e066c515ce

nickfarrow:
Apologies for the substantial mess this caused!

When making this PR it allowed *some* android target to build where I was having substantial trouble with ring. It has broken many other targets however and this was not readily apparent.

I think in the future I should take more time to test across other targets?

Big thanks for your work resolving this @thunderbiscuit

PR #135, opened 2024-06-26
Add use-rustls-ring feature

thunderbiscuit:
This PR adds the ability to build the client using the `ring` dependency for `rustls` instead of the new default `aws-lc-rs`.

As of the [`0.23.0` release](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0), rustls changed its default cryptography provider to [aws-lc-rs](https://crates.io/crates/aws-lc-rs). This new library is actually a set of bindings to a C library maintained by AWS, and they provide prebuilt bindings for [some platforms](https://aws.github.io/aws-lc-rs/platform_support.html) but not all. On these other platforms, the compilation step will attempt to build the bindings, requiring extra dependencies (CMake, libclang and others depending on the platform). This compilation step is what is currently breaking our Android and Swift builds for bdk-ffi. It is certainly possible to build the bindings (and the AWS docs on it are very nice), but for some reason I have not been able to make it work everywhere yet (local, CI, Windows).

This PR enables us to use the previous default `ring` library for rustls. I basically have to turn off the default features on `rustls` and re-enable all of them _except_ for the `aws_lc_rs`. We also have a few feature-gated constructs in the library, for which I needed to add the new proposed `use-rustls-ring` feature in order to make all of this work for us. Let me know if there are maybe better ways to achieve this!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ serde_json = { version = "^1.0" }
 
 # Optional dependencies
 openssl = { version = "0.10", optional = true }
-rustls = { version = "0.23", optional = true }
+rustls = { version = "0.23", optional = true, default-features = false }
 webpki-roots = { version = "0.25", optional = true }
 
 byteorder = { version = "1.0", optional = true }
@@ -41,5 +41,6 @@ default = ["proxy", "use-rustls"]
 minimal = []
 debug-calls = []
 proxy = ["byteorder", "winapi", "libc"]
-use-rustls = ["webpki-roots", "rustls"]
+use-rustls = ["webpki-roots", "rustls/default"]
+use-rustls-ring = ["webpki-roots", "rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12"]
 use-openssl = ["openssl"]
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -45,3 +45,4 @@ jobs:
       - run: cargo check --verbose --no-default-features --features=minimal,debug-calls
       - run: cargo check --verbose --no-default-features --features=proxy,use-openssl
       - run: cargo check --verbose --no-default-features --features=proxy,use-rustls
+      - run: cargo check --verbose --no-default-features --features=proxy,use-rustls-ring
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,14 +25,22 @@ extern crate log;
 #[cfg(feature = "use-openssl")]
 extern crate openssl;
 #[cfg(all(
-    any(feature = "default", feature = "use-rustls"),
+    any(
+        feature = "default",
+        feature = "use-rustls",
+        feature = "use-rustls-ring"
+    ),
     not(feature = "use-openssl")
 ))]
 extern crate rustls;
 extern crate serde;
 extern crate serde_json;
 
-#[cfg(any(feature = "use-rustls", feature = "default"))]
+#[cfg(any(
+    feature = "default",
+    feature = "use-rustls",
+    feature = "use-rustls-ring"
+))]
 extern crate webpki_roots;
 
 #[cfg(any(feature = "default", feature = "proxy"))]
@@ -51,7 +59,8 @@ mod batch;
 
 #[cfg(any(
     all(feature = "proxy", feature = "use-openssl"),
-    all(feature = "proxy", feature = "use-rustls")
+    all(feature = "proxy", feature = "use-rustls"),
+    all(feature = "proxy", feature = "use-rustls-ring")
 ))]
 pub mod client;
 
@@ -66,7 +75,8 @@ pub use api::ElectrumApi;
 pub use batch::Batch;
 #[cfg(any(
     all(feature = "proxy", feature = "use-openssl"),
-    all(feature = "proxy", feature = "use-rustls")
+    all(feature = "proxy", feature = "use-rustls"),
+    all(feature = "proxy", feature = "use-rustls-ring")
 ))]
 pub use client::*;
 pub use config::{Config, ConfigBuilder, Socks5Config};
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -23,7 +23,11 @@ use bitcoin::{Script, Txid};
 use openssl::ssl::{SslConnector, SslMethod, SslStream, SslVerifyMode};
 
 #[cfg(all(
-    any(feature = "default", feature = "use-rustls"),
+    any(
+        feature = "default",
+        feature = "use-rustls",
+        feature = "use-rustls-ring"
+    ),
     not(feature = "use-openssl")
 ))]
 use rustls::{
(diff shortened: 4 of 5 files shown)


oleonardolima (approved):
Concept ACK c7a96f7264348537f4b57dd149fd114f22844b16

Cargo.toml
```diff
@@ -41,5 +41,6 @@ default = ["proxy", "use-rustls"]
 minimal = []
 debug-calls = []
 proxy = ["byteorder", "winapi", "libc"]
-use-rustls = ["webpki-roots", "rustls"]
+use-rustls = ["webpki-roots", "rustls/default"]
+use-rustls-ring = ["proxy", "webpki-roots", "rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12"]
```
oleonardolima, line 45: @thunderbiscuit Do we need the proxy feature to be set when using `use-rustls-ring` ?
Is this just to keep the standard like the `default` ones ?

oleonardolima (commented):
I think the documentation here could be slightly updated to mention both options use-rustls and use-rusts-ring, wdyt ?

https://github.com/bitcoindevkit/rust-electrum-client/blob/c7a96f7264348537f4b57dd149fd114f22844b16/src/lib.rs#L6-L8

LLFourn:
Can we just expose the rustls provider API here: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html so I can pass in my own provider?
We can have a  `rustls-aws-lc` feature that adds a constructor that uses that. Can do the same for `rustls-ring`. But I'd like to able to fix this kind of issue in the future without having to wait for a merge.

oleonardolima:
> Can we just expose the rustls provider API here: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html so I can pass in my own provider? We can have a `rustls-aws-lc` feature that adds a constructor that uses that. Can do the same for `rustls-ring`. But I'd like to able to fix this kind of issue in the future without having to wait for a merge.

That's a good question! I'm not sure and would need to take a look at the docs and give it a shot, but if that's feasible, I agree that could be a better approach.

oleonardolima:
> > Can we just expose the rustls provider API here: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html so I can pass in my own provider? We can have a `rustls-aws-lc` feature that adds a constructor that uses that. Can do the same for `rustls-ring`. But I'd like to able to fix this kind of issue in the future without having to wait for a merge.
> 
> That's a good question! I'm not sure and would need to take a look at the docs and give it a shot, but if that's feasible, I agree that could be a better approach.

@thunderbiscuit Taking a look at the documentation it looks like really simple to follow Lloyd's suggestion.

You'd need to use the proper `.builder_with_provider()` on `ClientConfig`, probably on the lines:

https://github.com/bitcoindevkit/rust-electrum-client/blob/64c77ee1bc90c5e19ec0d401b97e420d495d7c6d/src/raw_client.rs#L387-L410

Cargo.toml
thunderbiscuit: Yep you're right. Fixed in my latest push.

thunderbiscuit:
@LLFourn @oleonardolima 
> Can we just expose the rustls provider API here: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html so I can pass in my own provider? We can have a rustls-aws-lc feature that adds a constructor that uses that. Can do the same for rustls-ring. But I'd like to able to fix this kind of issue in the future without having to wait for a merge.

This is interesting and might be a good way to go long term, but is outside the scope of this PR IMO, which is just a flag fix for rustls changing their default provider. I'll open an issue so we can tackle this approach in a different PR which should get its own review. 

The builds are currently breaking for 3 known projects, and I want to make sure we help them move forward. If this non-intrusive flag allows users to fix their build in the short run, let's get it in.

Unless there are oppositions to this, consider merging @notmandatory.

Interested parties:
@oleonardolima 
@ValuedMammal 
@LLFourn 
@notmandatory 
@praveenperera

oleonardolima (approved):
ACK b415b5cf826c799febeefa3c04ff2e093554f3ff
I do agree that's a good approach, and we can address the further improvement in another PR.

LLFourn:
ACK b415b5cf826c799febeefa3c04ff2e093554f3ff

LLFourn (approved):

Cargo.toml
```diff
@@ -41,5 +41,6 @@ default = ["proxy", "use-rustls"]
 minimal = []
 debug-calls = []
 proxy = ["byteorder", "winapi", "libc"]
-use-rustls = ["webpki-roots", "rustls"]
+use-rustls = ["webpki-roots", "rustls/default"]
+use-rustls-ring = ["rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12"]
```
praveenperera, line 45:
```suggestion
use-rustls-ring = ["rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12", "webpki-roots"]
```

praveenperera:
@thunderbiscuit got error.

44 | extern crate webpki_roots;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
use-rustls-ring does not enable the webpki_roots feature

use-rustls-ring = ["rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12"]
But it is imported here

#[cfg(any(
    feature = "default",
    feature = "use-rustls",
    feature = "use-rustls-ring"
))]
extern crate webpki_roots;
its working and compiling after adding the webpki_roots to the deps for the new feature: https://github.com/bitcoinppl/rust-electrum-client/commit/e3314aa5720653f110b579fe80085cc52e611189

oleonardolima:
> @thunderbiscuit got error.
> 
> ```
> 44 | extern crate webpki_roots;
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
> ```
> 
> `use-rustls-ring` does not enable the webpki_roots feature
> 
> ```toml
> use-rustls-ring = ["rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12"]
> ```
> 
> But it is imported here
> 
> ```rust
> #[cfg(any(
>     feature = "default",
>     feature = "use-rustls",
>     feature = "use-rustls-ring"
> ))]
> extern crate webpki_roots;
> ```
> 
> its working and compiling after adding the webpki_roots to the deps for the new feature: [bitcoinppl@e3314aa](https://github.com/bitcoinppl/rust-electrum-client/commit/e3314aa5720653f110b579fe80085cc52e611189)

Yes, you're right it's missing as it's being used here:

https://github.com/bitcoindevkit/rust-electrum-client/blob/64c77ee1bc90c5e19ec0d401b97e420d495d7c6d/src/raw_client.rs#L392-L399

It got my attention now, that our CI is not properly covering these tests. 🤔

Cargo.toml
thunderbiscuit: Ah thanks for catching this 👍

thunderbiscuit:
Note that I updated the CI to run the `use-rustls-ring` feature. Thanks @oleonardolima!

oleonardolima (approved):
ACK 8d71f9597c99d3228bfab2643ddccfc0f44addca

notmandatory (approved):
ACK 8d71f9597c99d3228bfab2643ddccfc0f44addca

PR #150, opened 2024-09-27
fix(insecure-tls): `NoCertificateVerification` implementation

### LINKED ISSUE #149: fix+refactor(tls): it should properly handle self-signed certificates, return a clear error otherwise
**Describe the enhancement**  

It has been noticed both in CI tests, which currently rely on blockstream's electrum client, and by other users in https://github.com/bitcoindevkit/bdk/issues/1598 and https://github.com/wizardsardine/liana/issues/1300 that the `electrum-client` does not properly connect to electrum servers with self-signed certificates while using the `validate_domain: false` settings, and neither returns a proper and clear regarding the problem. There is some issue when using the `rustls` crate that it fails for self-signed certificates, on other hand `openssl` works just fine.

Therefore, I'm creating this issue mainly for two purposes:

1. Improve the documentation regarding the usage of `validate_domain: false`, when using either `openssl` and `rustls` with it's expected behavior.
2. Improve the error handling and propagation, reporting proper TLS certificate validation errors to the user.
3. Investigate and fix the inner issue with `rustls` custom certificate validation.

**Use case**  
 

Allow users to properly use and connect electrum servers with self-signed certificates, either with `openssl` or `rustls`.

**Additional context**
 

https://github.com/rustls/rustls/issues/124
https://github.com/lightningnetwork/lnd/issues/5450
https://github.com/rigelminer/rigel/issues/130

oleonardolima:
fixes #149 https://github.com/bitcoindevkit/bdk/issues/1598

### Description

It has been noticed some issues by both users and developers, as reported in #149, https://github.com/bitcoindevkit/bdk/issues/1598 and https://github.com/wizardsardine/liana/issues/1300, when using the library with `use-rustls-{ring}` feature to connect to electrum servers that use self-signed certificates, there are even some issues when trying to connect to `ssl://electrum.blockstream.info:50002` server. 

To connect in an insecure manner either with `rustls` or `openssl` features, the user can set the `validate_domain` field in the `Config` to false, this will either set the `SslVerifyMode::NONE` when using `openssl`, or use the custom `NoCertificateVerification` for the 
`rustls::client::danger::ServerCertVerifier` trait when using `rustls`, that said it should ignore the certificate verification when used.

At the current library state, it's failing because we didn't set up the supported `rustls::SignatureScheme` properly, returning an empty vector at the moment. This PR focuses on fixing this issue by relying on the `CryptoProvider` in usage to get the correct and supported signature schemes.

As part of the research to understand the problem, I've noticed that ideally, we should still use both the `rustls::webpki::verify_tls12_signature` and `rustls::webpki::verify_tls12_signature` and only rely on `rustls::client::danger::ServerCertVerified::assertion()` to ignore the certificate verification, however, it would still fail in scenarios such as https://github.com/bitcoindevkit/bdk/issues/1598 which uses X.509 certificates with any version other than 3 (it uses version 1), see [here](https://github.com/rustls/webpki/blob/1a0d1646d0bb1b7851bf81c6244cab09c352d8ef/src/cert.rs#L202-L218).

I kept the current behavior to also ignore the TLS signature, but I still would like to bring this to the discussion, should we validate it properly and update the documentation to mention the `webpki` limitation instead ?

### Notes to the reviewers

I kept the current behavior to also ignore the TLS signature, but I still would like to bring this to the discussion, should we validate it properly and update the documentation to mention the `webpki` limitation instead ?

### Changelog notice

- Updates the `NoCertificateVerification` implementation for the 
`rustls::client::danger::ServerCertVerifier` to use the `rustls::SignatureScheme` from `CryptoProvider` in use.

#### All Submissions:

* [x] I've signed all my commits
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -299,23 +299,29 @@ impl RawClient<ElectrumSslStream> {
 ))]
 mod danger {
     use crate::raw_client::ServerName;
-    use rustls::client::danger::ServerCertVerified;
-    use rustls::pki_types::CertificateDer;
-    use rustls::pki_types::UnixTime;
-    use rustls::Error;
+    use rustls::client::danger::{HandshakeSignatureValid, ServerCertVerified};
+    use rustls::crypto::CryptoProvider;
+    use rustls::pki_types::{CertificateDer, UnixTime};
+    use rustls::DigitallySignedStruct;
 
     #[derive(Debug)]
-    pub struct NoCertificateVerification {}
+    pub struct NoCertificateVerification(CryptoProvider);
+
+    impl NoCertificateVerification {
+        pub fn new(provider: CryptoProvider) -> Self {
+            Self(provider)
+        }
+    }
 
     impl rustls::client::danger::ServerCertVerifier for NoCertificateVerification {
         fn verify_server_cert(
             &self,
-            _end_entity: &CertificateDer,
-            _intermediates: &[CertificateDer],
-            _server_name: &ServerName,
-            _ocsp_response: &[u8],
+            _end_entity: &CertificateDer<'_>,
+            _intermediates: &[CertificateDer<'_>],
+            _server_name: &ServerName<'_>,
+            _ocsp: &[u8],
             _now: UnixTime,
-        ) -> Result<ServerCertVerified, Error> {
+        ) -> Result<ServerCertVerified, rustls::Error> {
             Ok(ServerCertVerified::assertion())
         }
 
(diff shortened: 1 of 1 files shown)


oleonardolima:
I also think we should update `validate_domain` to something like `insecure` instead, but would like to know what others think about it.

LLFourn:
I didn't really understand the nuance here but we should validate domains by default and then have an opt out I think.

ValuedMammal (approved):
ACK 05771a81d7d0383a089fc20f1e1e228202709f01

This looks to be an improvement over what was there before and it fixes an issue that prevented connecting to an electrum server with a self-signed cert using rustls. I tested setting `validate_domain = false` and was able to get a response from each of blockstream's esplora, the server at "ssl://testnet.aranguren.org:51002" (fulcrum), and an electrs instance on my local network.

oleonardolima:
> I didn't really understand the nuance here but we should validate domains by default and then have an opt out I think.

In the previous state, when using `rustls` for TLS in an insecure manner (explicitly called/opted out by the user through the `validate_domain` configuration) it wasn't properly setting up the implementation for `ServerCertVerifier`, more specifically on the validation for the supported schemes.

> This looks to be an improvement over what was there before and it fixes an issue that prevented connecting to an electrum server with a self-signed cert using rustls. I tested setting validate_domain = false and was able to get a response from each of blockstream's esplora, the server at "ssl://testnet.aranguren.org:51002" (fulcrum), and an electrs instance on my local network.

Ideally, we SHOULD properly implement the verification on both `verify_tls12_signature` and `verify_tls12_signature`, it should still work fine with self-signed certificates, in an insecure manner, but that doesn't work with all self-signed certificates, such as `ssl://testnet.aranguren.org:51002`, as this one and some other uses a certificate version other than 3.

Do you think it's fine to have this tradeoff? And, any thoughts on changing the parameter to `insecure` too?

notmandatory (approved):
ACK 05771a81d7d0383a089fc20f1e1e228202709f01

LLFourn:
ACK 05771a81d7d0383a089fc20f1e1e228202709f01

I think we should change how this works. People should not be using `validate_domain = false` except for testnet. We should have an API to add certificates. Better yet add a TOFU API to give you a certificate to persist when you connect to a server for the first time. I'm not sure we should even use webpki here. There aren't that many public electrum servers. We can keep a list updated in the client itself.

PR #158, opened 2024-11-29
chore: Enforce min `rustls` version to 0.23.19

evanlinjin:
`rustls` versions 0.23.18 and 0.23.19 contains fix for vulnerability RUSTSEC-2024-0399. However, 0.23.18 bumps MSRV to 1.71. 0.23.19 reverts MSRV back to 1.63.

We enforce min `rustls` version to 0.23.19 to make it easier to compile on MSRV and ensure we include the RUSTSEC-2024-0399 fix.

Note that in CI, I decided to pin `rustls` dependency to 0.23.19 explicitly. This is because in future versions of `rustls`, the MSRV will be changed to 1.71.

Context: https://github.com/rustls/rustls/pull/2244

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Pin dependencies for MSRV
         if: matrix.rust == '1.63.0'
         run: |
-          cargo update -p rustls --precise "0.23.17"
+          cargo update -p rustls --precise "0.23.19"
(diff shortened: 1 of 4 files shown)


notmandatory (approved):
ACK ed0cdae32971fe7b40d5587a55b2299fe74d554f

Thanks for fixing this up, I went ahead and updated the README and bumped the patch version so it's ready to release.

Issue #75, opened 2022-07-29
Transactions being returned out of order


0xarvin:
Via BDK, we're pulling transactions for a Wallet through a signet electrum instance. Recently, we've noticed that the order that we call [batch_transaction_get_raw](https://github.com/bitcoindevkit/rust-electrum-client/blame/866c2424e1ce3dff4daf1abb4daf5d470dda2abf/src/raw_client.rs#L932) isn't being respected by electrum and results in the transactions being returned by the client to also be out of order as [the reader takes the top result if it's unable to grab a hold of the lock on the reader](https://github.com/bitcoindevkit/rust-electrum-client/blame/866c2424e1ce3dff4daf1abb4daf5d470dda2abf/src/raw_client.rs#L606-L630)

This doesn't affect production calls in BDK but only in tests as there's a [debug assert here](https://github.com/bitcoindevkit/bdk/blob/73d4f6d3b1549df2a2f8a61df1f97de3dcf9f6f5/src/blockchain/electrum.rs#L276). Was wondering if there's anything that I'm missing or if this is a known issue.

Wallet Descriptor for example: 
`wsh(sortedmulti(2,[e5cd860e/84'/1'/0']tpubDFVZtuMvX7n6DCWbNAgq361KiJdVFBy4D1sZmdafF1zDNW6dv79MwbK6JBRxGV42BS4STayPsG7Prn3QT7LQwzfaBhE8YhgCs4VzhkYVWTg/0/*,[73ee041b/84'/1'/0']tpubDEPCQp4KHW3sS5o18QbknuC6dtQJPFgWtS6e8wm6ewnc2B1vucCnKf9uf9gCddHeJoK5d5hgJv5ji5Q8VnpT4h9ydsi9FE4t8F2GoPrFPYa/0/*,[08a73043/84'/1'/0']tpubDEZ565bfePfqSVkSfrLaR7g5nruLDWxJFKkJvKWUpMEQBsS9p2f5um5uTwgmKWuHbyoiythmEvDRSgdEoYBgLXfosBsjMBTp1jev9wfi2q9/0/*`

Network: `Signet`

afilini:
I don't think this was a known issue, and I'm guessing this may be related to the specific server your are running because I've never seen this happen with electrs.

Looking at the code it's also very weird that this is happening: we do have code to re-order the replies from the server for batch requests:

[...]

PR #160, opened 2025-01-22
fix: fix batch ordering issue

marshallyale:
Fixes issue #75 
This is my first open source pull request so I apologize for any formatting issues. Additionally, I don't know the repository as well as others so there may be a better way to implement the fix.

I believe I found the root cause of this. I added a pull request to fix, but I'm going to copy/paste what I believe is causing the error.

The main issue in the code is inside raw_client.rs inside the `recv` method implementation (snippet below):
https://github.com/bitcoindevkit/rust-electrum-client/blob/805ea0af307e6465f23c2d7f25a32d7ff61fe7ec/src/raw_client.rs#L671-L685

When this is first called, the `self._reader_thread` will run. Inside the `self._reader_thread`, if the request id matches the response id, everything works fine. However, if the request id does not match the response id, we run the following code:
https://github.com/bitcoindevkit/rust-electrum-client/blob/805ea0af307e6465f23c2d7f25a32d7ff61fe7ec/src/raw_client.rs#L602-L612
 
The channel that the response is sent back into is not unique, but rather all the channels share the same sender.clone() and receiver. The only validation that is done is to check that the request id is still being searched for inside `self.waiting_map`. This means that the receiver channel receives whatever the next response is into the channel without any validation that it matches the request id which happens here `match receiver.recv()?`.

This is fixed by implementing unique channels for every request id. This fix can be verified with the code johnzweng used to show the issue

If you run this with the initial code, it will error out after 1-10 cycles normally. However, after the fix this runs indefinitely.

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -1190,6 +1190,26 @@ mod test {
         assert_eq!(resp.hash_function, Some("sha256".into()));
         assert_eq!(resp.pruning, None);
     }
+
+    #[test]
+    #[ignore = "depends on a live server"]
+    fn test_batch_response_ordering() {
+        // The electrum.blockstream.info:50001 node always sends back ordered responses which will make this always pass.
+        // However, many servers do not, so we use one of those servers for this test.
+        let client = RawClient::new("exs.dyshek.org:50001", None).unwrap();
+        let heights: Vec<u32> = vec![1, 4, 8, 12, 222, 6666, 12];
+        let result_times = [
+            1231469665, 1231470988, 1231472743, 1231474888, 1231770653, 1236456633, 1231474888,
+        ];
+        // Check ordering 10 times. This usually fails within 5 if ordering is incorrect.
+        for _ in 0..10 {
+            let results = client.batch_block_header(&heights).unwrap();
+            for (index, result) in results.iter().enumerate() {
+                assert_eq!(result_times[index], result.time);
+            }
+        }
+    }
+
     #[test]
     fn test_relay_fee() {
         let client = RawClient::new(get_test_server(), None).unwrap();
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -3,7 +3,7 @@
 //! This module contains the definition of the raw client that wraps the transport method
 
 use std::borrow::Borrow;
-use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
+use std::collections::{BTreeMap, HashMap, VecDeque};
 use std::io::{BufRead, BufReader, Read, Write};
 use std::mem::drop;
 use std::net::{TcpStream, ToSocketAddrs};
@@ -539,11 +539,10 @@ impl<S: Read + Write> RawClient<S> {
 
                 if let Some(until_message) = until_message {
                     // If we are trying to start a reader thread but the corresponding sender is
-                    // missing from the map, exit immediately. This can happen with batch calls,
-                    // since the sender is shared for all the individual queries in a call. We
-                    // might have already received a response for that id, but we don't know it
-                    // yet. Exiting here forces the calling code to fallback to the sender-receiver
-                    // method, and it should find a message there waiting for it.
+                    // missing from the map, exit immediately. We might have already received a
+                    // response for that id, but we don't know it yet. Exiting here forces the
+                    // calling code to fallback to the sender-receiver method, and it should find
+                    // a message there waiting for it.
                     if self.waiting_map.lock()?.get(&until_message).is_none() {
                         return Err(Error::CouldntLockReader);
                     }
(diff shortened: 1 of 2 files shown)


marshallyale:
No worries, I should have just rebased and pushed this. I'm working off my forked repo currently so there's no rush on getting any changes synced. Let me know if there's anything else I need to change!

oleonardolima (commented):
Thanks for working on this fix!

I've reviewed the issue and proposed changes, it seems it's a library problem as it's not management and matching context between requests and responses, as the response for a batch call may be out of ordered (as per JSON-RPC spec).

Using an exclusive channel for each request in a batch, solves the problem, however I'm trying to think about any shortcomings of that approach.

oleonardolima:
I'm taking another look at this one this week, and I think it's best to come up with two `#[test]`'s: (i) checking/asserting that the request_id's are correct within a batch request, and  (ii) single requests / batch requests doesn't mixups the request_id's too. By having a test running on CI/locally it's easier to iterate over this solution and other ones.

ValuedMammal:
I think the unique channels approach in this PR is a valid solution. Perhaps an alternative would be to change `ChannelMessage::Response` to carry the response id in a tuple of `(usize, Value)` and use this when inserting into `answers` during `batch_call`. Since `answers` is an ordered map, we'll still return the responses in the correct order.

notmandatory:
@marshallyale did you see the comment above? If you don't have time to look at it now we may open a new PR to try out his suggestion.

marshallyale:
Yes, apologies for the delay. I will write some tests and look into the ChannelMessage::Response this weekend sometime.

marshallyale:
I've added a single test that tests batch_block_header and asserts that the timestamp matches up to the expected times. This is looped 10 times, although it would fail within 5 times normally. I had to use a different client (exs.dyshek.org:50001) since it appears that electrum.blockstream.info always returns results in order, which is not true of other servers. 

I looked at trying to implement tests comparing request ids to response ids, but there isn't a great way to do this without doing a bit of code refactor and adding additional loops to the code. Both batch_call and raw_call don't return the json response ids, so we would have to break the actual json responses into a separate function that returns the raw request-response matchup, and then iterate over this to get the actual result in its current format.

Additionally, I looked into changing ChannelMessage::Response from Value to (req_id, value). This kind of gets back to the original issue with each request sharing the same channel for receiving, as the receiver will receive the response but would then have to check that the response id is equal to the request id and if not, would have to put this message back in the channel until the correct request id gets it. It could definitely be done, but would require a bit more reworking than the current separate channels approach.

I think that this could probably be done in a more elegant/faster way, but would require more of an overhaul of the send/receive logic in the code. Maybe this is just a function of me being very new to the code, but I personally think the recv function and _reader_thread code is confusing which makes it hard to come up with alternatives without a major code refactor.

Finally, it looks like the tests failed because actions/cache@v2 is deprecated. Want me to update to v3 (or v4)?

src/raw_client.rs
```diff
@@ -762,22 +762,23 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
 
         for (method, params) in batch.iter() {
             let req = Request::new_id(
                 self.last_id.fetch_add(1, Ordering::SeqCst),
                 method,
                 params.to_vec(),
             );
-            missing_responses.insert(req.id);
+            // Add distinct channel to each request so when we remove our request id (and sender) from the waiting_map
+            // we can be sure that the response gets sent to the correct channel in self.recv
```
ValuedMammal, line 777: nit: I want to update the stale comment on L543 that still refers to the "shared sender".

src/raw_client.rs
```diff
@@ -1190,6 +1191,25 @@ mod test {
         assert_eq!(resp.hash_function, Some("sha256".into()));
         assert_eq!(resp.pruning, None);
     }
+
+    #[test]
```
ValuedMammal, line 1195: Thanks for adding this. I haven't vetted the reliability of this server and so would prefer to ignore it for unit tests.

```suggestion
    #[test]
    #[ignore = "depends on a live server"]
```

ValuedMammal (changes requested):
Approach ACK

ValuedMammal:
> Finally, it looks like the tests failed because actions/cache@v2 is deprecated. Want me to update to v3 (or v4)?

If you want to push an update to `actions/cache` that would be fine or else we can do it in another PR (#167).

src/raw_client.rs
marshallyale: Removed the shared sender part of the comment on L543

src/raw_client.rs
marshallyale: Added the ignore statement into the latest commit.

ValuedMammal (approved):
ACK b6230be3b47d65d4866d466bda36322637957e01

ValuedMammal (approved):
reACK a3488f4ff9007f0682f8e08cdedb6302498ac190

Thanks again @marshallyale

PR #163, opened 2025-02-07
Allow types that reference `ElectrumApi` to also implement `ElectrumApi`

evanlinjin:
This is an alternative and non-breaking alternative to https://github.com/bitcoindevkit/bdk/pull/1820.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -2,12 +2,174 @@
 
 use std::borrow::Borrow;
 use std::convert::TryInto;
+use std::ops::Deref;
 
 use bitcoin::consensus::encode::{deserialize, serialize};
 use bitcoin::{block, Script, Transaction, Txid};
 
 use crate::batch::Batch;
 use crate::types::*;
+
+impl<E: Deref> ElectrumApi for E
+where
+    E::Target: ElectrumApi,
+{
+    fn raw_call(
+        &self,
+        method_name: &str,
+        params: impl IntoIterator<Item = Param>,
+    ) -> Result<serde_json::Value, Error> {
+        (**self).raw_call(method_name, params)
+    }
+
+    fn batch_call(&self, batch: &Batch) -> Result<Vec<serde_json::Value>, Error> {
+        (**self).batch_call(batch)
+    }
+
+    fn block_headers_subscribe_raw(&self) -> Result<RawHeaderNotification, Error> {
+        (**self).block_headers_subscribe_raw()
+    }
+
+    fn block_headers_pop_raw(&self) -> Result<Option<RawHeaderNotification>, Error> {
+        (**self).block_headers_pop_raw()
+    }
+
+    fn block_header_raw(&self, height: usize) -> Result<Vec<u8>, Error> {
+        (**self).block_header_raw(height)
+    }
+
+    fn block_headers(&self, start_height: usize, count: usize) -> Result<GetHeadersRes, Error> {
+        (**self).block_headers(start_height, count)
+    }
+
+    fn estimate_fee(&self, number: usize) -> Result<f64, Error> {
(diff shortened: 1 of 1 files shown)


nymius (commented):
It looks convoluted, but works.
I did the following analysis for the `call_made` function to remove some of the ergonomics Rust adds on top of the language. Each line is equivalent to the next one. I hope it helps someone:
```rust
    (**self).calls_made()
    (*(*&self)).calls_made()
    (*self).calls_made()
    (*Deref::deref(&self)).calls_made()
    (*&Self::Target).calls_made()
    Self::Target.calls_made()
```

From the read of bitcoindevkit/bdk#1820 and bitcoindevkit/bdk#1825, I got the following assumptions:

1. This will open up the possibility to have multiple `BdkElectrumClient` consuming from the same client.
2. The requirement of implementing `ElectrumApi` for `Deref<Target = ElectrumApi>` or to take `Borrow<ElectrumApi` for `BdkElectrumClient` inner is to respect the use of the trait API, and not allow something like:
```rust
struct RandomClient {
    client: electrum_client::Client,
}

impl RandomClient {
    pub fn new(e: electrum_client::Client) -> Self {
      RandomClient {
        client: e
      }
    }

    pub fn just_this_method(&self) -> u64 { 0 }
}

impl<T: ElectrumApi> Deref for RandomClient<T> {
  type Target = T;

  fn deref(&self) -> &Self::Target {
    &self.client
  }
}

let client = BdkElectrumClient::new(RandomClient::new(electrum_client::Client::new(ELECTRUM_URL)?));
```

which would fulfill just requiring `BdkElectrumClient::inner` to be `Deref<Target = ElectrumApi>` but it won't behave like `ElectrumApi`.

As a last remark, notice that this will also be true:
```rust
is_impl::<Arc<Arc<Arc<Arc<Arc<Arc<Arc<FakeApi>>>>>>>>();
```

evanlinjin:
@nymius what do you mean by:

> The requirement of implementing ElectrumApi for Deref<Target = ElectrumApi> or to take Borrow<ElectrumApi for BdkElectrumClient inner is to respect the use of the trait API, and not allow something like:

nymius:
It is how I rationalized [your comment](https://github.com/bitcoindevkit/bdk/pull/1820#issuecomment-2644446744) on the original bitcoindevkit/bdk#1820 PR. I understood the original solution was too "loose" with the inner type, allowing structs that didn't implement `ElectrumApi` directly (like `RandomClient` above) or requiring types which already implemented `ElectrumApi` originally to also implement `Deref<Target = ElectrumApi>` (maybe you were only concerned about the last one).

ValuedMammal (approved):
ACK eb38c038b205c2c7bc21299ccf62cc325392afc8

evanlinjin:
@nymius The original solution was too "tight" in the sense that the changes forced us to only have something that derefs to `impl ElectrumApi`. In other words, something just `impl ElectrumApi` was not enough.

It loosened it in the sense that we can have smart pointers and references to `impl ElectrumApi` (which would be nice). However, we still need the old behavior of taking in something owned and the changes broke that.

nymius:
> @nymius The original solution was too "tight" in the sense that the changes forced us to only have something that derefs to `impl ElectrumApi`. In other words, something just `impl ElectrumApi` was not enough.
> 
> It loosened it in the sense that we can have smart pointers and references to `impl ElectrumApi` (which would be nice). However, we still need the old behavior of taking in something owned and the changes broke that.

I see, I used terms poorly, but your explanation is very clear and concise, thanks!

LLFourn:
ACK eb38c038b205c2c7bc21299ccf62cc325392afc8

PR #164, opened 2025-02-09
`relay_fee` result should be strongly typed

rustaceanrob:
From the docs:

> Returns the minimum accepted fee by the server’s node in Bitcoin, not Satoshi.

That last nuance is irrelevant if the API just returns the `Amount` type. Same thing should be done for `estimate_fee` to use `FeeRate`, but that one is non-trivial.


src/raw_client.rs
```diff
@@ -879,17 +879,18 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
             .ok_or_else(|| Error::InvalidResponse(result.clone()))
     }
 
-    fn relay_fee(&self) -> Result<f64, Error> {
+    fn relay_fee(&self) -> Result<Amount, Error> {
```
ValuedMammal, line 882: I know the docs aren't clear in this regard but I think it would make sense for this to return a `FeeRate` which I believe was the approach in #136 also.

rustaceanrob: I don't know the electrum protocol, my interpretation of the docs was this is an _absolute_ fee that a transaction must have, so an `Amount` is the fee required to broadcast with that node.

oleonardolima: I also interpret the docs as this one being an absolute fee, but some validation on electrum code besides documentation is needed.

ValuedMammal: To be clear I'm referring to "relay fee" as described in `getnetworkinfo` RPC if indeed that's what this method refers to. https://bitcoincore.org/en/doc/28.0.0/rpc/network/getnetworkinfo/

I agree that the value can be parsed as an amount in BTC (same for `estimate_fee`), but that still leaves #136 unresolved where the user has to interpret the value as a useful feerate.

> this is an absolute fee that a transaction must have, so an Amount is the fee required to broadcast with that node

If you consider `test_relay_fee` below, that would be saying that the blockstream server refuses to relay anything that pays less than 1000 sats in fees which I don't think is the case.

src/api.rs
```diff
@@ -97,8 +97,8 @@ pub trait ElectrumApi {
     /// Estimates the fee required in **Bitcoin per kilobyte** to confirm a transaction in `number` blocks.
     fn estimate_fee(&self, number: usize) -> Result<f64, Error>;
 
-    /// Returns the minimum accepted fee by the server's node in **Bitcoin, not Satoshi**.
-    fn relay_fee(&self) -> Result<f64, Error>;
+    /// Returns the minimum accepted fee by the server's node.
+    fn relay_fee(&self) -> Result<Amount, Error>;
```
ValuedMammal, line 101:
```suggestion
    /// Returns the minimum accepted fee by the server's node per kvB.
    fn relay_fee(&self) -> Result<Amount, Error>;
```

src/raw_client.rs
oleonardolima: At least it seems both romanz/electrs and blockstream/electrs rely on the RPC you mentioned, basically propagating the `Amount`/`f64`, maybe a FeeRate really does fit it better.

ValuedMammal:
+1 for using Amount for `relay_fee`. But I think FeeRate is also acceptable.

Issue #137, opened 2024-07-27
Allow passing custom crypto provider


thunderbiscuit:
See [comment here](https://github.com/bitcoindevkit/rust-electrum-client/pull/135#issuecomment-2244180324).

PR #171, opened 2025-07-08
Install `rustls`'s `CryptoProvider` based on features

tnull:
Previously, we'd already assume `use-rustls` to use the default `aws-lc-rs` provider and `use-rustls-ring` to use the `ring` `CryptoProvider`, e.g., for `NoCertificateVerification`.

However, we **wouldn't** actually install the respective provider based on the features, leading to a **reachable** panic at runtime when user tried to access `ssl://` Electrum servers.

Here, we fix this omission and install the default provider according to the configured features.

(cc @oleonardolima @thunderbiscuit)

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -406,6 +406,29 @@ impl RawClient<ElectrumSslStream> {
     ) -> Result<Self, Error> {
         use std::convert::TryFrom;
 
+        if rustls::crypto::CryptoProvider::get_default().is_none() {
+            // We install a crypto provider depending on the set feature.
+            #[cfg(feature = "use-rustls")]
+            rustls::crypto::CryptoProvider::install_default(
+                rustls::crypto::aws_lc_rs::default_provider(),
+            )
+            .map_err(|_| {
+                Error::CouldNotCreateConnection(rustls::Error::General(
+                    "Failed to install CryptoProvider".to_string(),
+                ))
+            })?;
+
+            #[cfg(feature = "use-rustls-ring")]
+            rustls::crypto::CryptoProvider::install_default(
+                rustls::crypto::ring::default_provider(),
+            )
+            .map_err(|_| {
+                Error::CouldNotCreateConnection(rustls::Error::General(
+                    "Failed to install CryptoProvider".to_string(),
+                ))
+            })?;
+        }
+
         let builder = ClientConfig::builder();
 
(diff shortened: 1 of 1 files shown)


oleonardolima (commented):
cACK 41c52285d6b325c0331b1a5e15d1a302874b7c28

I wasn't able to reproduce the panic locally. Do you have a specific scenario/impl that you could share?

I assume you're referring to the following panic: `no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point`

That's produced by [`rustls::ClientConfig::get_default_or_install_from_crate_features`](https://github.com/rustls/rustls/blob/d1bd2c86341eb491921feca3c9408061eb0262fe/rustls/src/crypto/mod.rs#L238-L253), is that right ?

It tries to install the default `CryptoProvider` for the given process, and I wonder why it's not able to do it in your scenario, are you using ambiguous features ?

I don't see much problem with installing it explicitly based on the features, as it's also recommended by `rustls` docs.

oleonardolima:
> > I don't see much problem with installing it explicitly based on the features, as it's also recommended by `rustls` docs.
> 
> Well, there is a slight problem in that it's recommended for _binaries_, while they suggest libraries to keep it open. However, as mentioned in the OP we already infringe on that for the `NoCertificateVerification`, so might as well do it for the cert-verifying path, at least until we find a better solution.

Indeed, it seems like an okay fix for now, but ideally, we should move forward towards #137 and expose `CryptoProvider` api.

thunderbiscuit (approved):
ACK 41c52285d6b325c0331b1a5e15d1a302874b7c28.

Seems like a decent fix until we actually do #137 (or we migrate to the new Electrum client? Not sure where that's at).

In any case I see no reason to hold this one up.

LagginTimes (approved):
ACK 41c52285d6b325c0331b1a5e15d1a302874b7c28

ValuedMammal:
`CryptoProvider::install_default` "can be called successfully at most once in any process execution", so I guess installing a crypto provider here should be conditional on whether `CryptoProvider::get_default().is_none()`. What do you think @tnull ?

ValuedMammal (approved):
ACK f24458248aa3265725958f4a1b25e113678cb056

oleonardolima (approved):
ACK f24458248aa3265725958f4a1b25e113678cb056

PR #182, opened 2025-10-31
Simplify cargo features

### LINKED ISSUE #42: Rename `use-rustls` and `use-openssl` features
According to Rust API guidelines feature names with `use-` prefix considered wrong: https://rust-lang.github.io/api-guidelines/naming.html#c-feature

This is justified, since ppl looking for features will frequently use `rustls` instead of `use-rustls` and will report issues - we had plenty of them in `rust-bitcoin` with similar `serde` situation.

There is two options:
1. Rename crates with `rustls_crate = { package = "rustls" }` + `extern crate rustls_crate as rustls` in `lib.rs` and rename features into `rustls`
2. Rename features into `tls` and `ssl`

I am pro second option

### LINKED ISSUE #91: Client is only exported when `proxy` feature is enabled.

ValuedMammal:
## Description

This PR is meant to streamline the handling of feature flags in the library, in particular by removing the Client's dependency on the `proxy` feature and overall simplifying the feature gating logic. Additionally, we move away from the old `use-*` feature naming convention by introducing simpler feature names, remove the unused `minimal` feature, and get rid of conditional compilation for `debug-calls` while retaining the ability to atomically count calls made.

fixes #91
fixes #42

### DIFF
--- a/justfile
+++ b/justfile
@@ -14,12 +14,22 @@ build:
 # Check code: formatting, compilation, linting, doc comments, and commit signature
 check:
    cargo +nightly fmt --all -- --check
-   cargo check --all-features --all-targets
+   @just _check-features
    cargo clippy --all-features --all-targets -- -D warnings
-   RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps
+   cargo rustdoc --all-features -- -D warnings
    @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
        echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
        true
+
+# Check feature configurations (matches CI)
+_check-features:
+   cargo check --features "default"
+   cargo check --no-default-features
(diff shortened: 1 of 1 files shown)


oleonardolima (commented):
utACK 51cfd778c2711403fa08706345ff55def588edf5

ValuedMammal:
I left the test job the same in that the tests are run using `--all-features`, then added a separate check job that only does `cargo check`. Let me know if you had something different in mind.

PR #188, opened 2025-12-03
Include latest error in AllAttemptsErrored

### LINKED ISSUE #186: Routinely returns `AllAttemptsErrored([])` on failed requests
In these parts of the code we returning **before** adding the latest error to the `errors: Vec<_>`. This causes this code to return `AllAttemptsErrored([])` which is a bit frustrating.

https://github.com/bitcoindevkit/rust-electrum-client/blob/761796c94ae29a41651914796c926bbabf8784c8/src/client.rs#L56-L65
https://github.com/bitcoindevkit/rust-electrum-client/blob/761796c94ae29a41651914796c926bbabf8784c8/src/client.rs#L80-L89

nabijaczleweli:
```
$ git grep AllAttemptsErrored
src/client.rs:                        return Err(Error::AllAttemptsErrored(errors));
src/client.rs:                                        return Err(Error::AllAttemptsErrored(errors));
```
fixed here
```
src/raw_client.rs:    Err(Error::AllAttemptsErrored(errors))
```
already correctly collected
```
src/types.rs:    AllAttemptsErrored(Vec<Error>),
src/types.rs:            Error::AllAttemptsErrored(errors) => {
```
consumers

Closes #186

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -38,63 +38,97 @@ pub struct Client {
 macro_rules! impl_inner_call {
     ( $self:expr, $name:ident $(, $args:expr)* ) => {
     {
-        let mut errors = vec![];
-        loop {
-            let read_client = $self.client_type.read().unwrap();
-            let res = match &*read_client {
+        impl_inner_call_impl($self, || {
+            match &*$self.client_type.read().unwrap() {
                 ClientType::TCP(inner) => inner.$name( $($args, )* ),
                 ClientType::SSL(inner) => inner.$name( $($args, )* ),
                 ClientType::Socks5(inner) => inner.$name( $($args, )* ),
-            };
-            drop(read_client);
-            match res {
-                Ok(val) => return Ok(val),
-                Err(Error::Protocol(_) | Error::AlreadySubscribed(_)) => {
-                    return res;
-                },
+            }
+        })
+    }}
+}
+
+fn impl_inner_call_impl<T>(
+    self_: &Client,
+    mut f: impl FnMut() -> Result<T, Error>,
+) -> Result<T, Error> {
+    let mut errors = vec![];
+    loop {
+        match f() {
+            Ok(val) => return Ok(val),
+            res @ Err(Error::Protocol(_) | Error::AlreadySubscribed(_)) => {
+                return res;
+            }
+            Err(e) => impl_inner_call_impl_err(self_, &mut errors, e)?,
+        }
+    }
+}
+
+fn impl_inner_call_impl_err(
+    self_: &Client,
+    errors: &mut Vec<Error>,
+    e: Error,
+) -> Result<(), Error> {
+    let failed_attempts = errors.len() + 1;
+
+    warn!(
+        "call '{}' failed with {}, retry: {}/{}",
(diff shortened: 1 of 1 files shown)


ValuedMammal:
In 59c28bd: `Include latest error in AllAttemptsErrored`

I understand that we can expect `errors` to be non-empty after pushing an error into it, however this code is dangerous and error prone due to the unchecked container access. The only time `e` is used outside of the `errors` vector is to pass a reference to the `warn!` macro, so it'd arguably be better to log the warning first before moving the value into `errors`. This is the suggestion I tried to convey in an earlier comment https://github.com/bitcoindevkit/rust-electrum-client/pull/188#pullrequestreview-3548676446.

Please also include a PR description that others can make sense of, and consider using conventional commits. Thanks for your contribution. @nabijaczleweli

nabijaczleweli:
Sure, whatever; applied.

You're free to rephrase the PR and commit descriptions as you see fit, from my POV "Include latest error in AllAttemptsErrored" follows all the relevant conventions and is human-readable.

ValuedMammal:
I had an additional comment related to the order of the logging statements. I think I would prefer this way, so that in the case of a failed attempt the latest error `e` can still be captured by the logs. Let me know what you think. @nabijaczleweli

```diff
--- a/src/client.rs
+++ b/src/client.rs
@@ -55,16 +55,15 @@ macro_rules! impl_inner_call {
                 Err(e) => {
                     let failed_attempts = errors.len() + 1;
 
+                    warn!("call '{}' failed with {}, retry: {}/{}", stringify!($name), e, failed_attempts, $self.config.retry());
+
+                    errors.push(e);
+
                     if retries_exhausted(failed_attempts, $self.config.retry()) {
                         warn!("call '{}' failed after {} attempts", stringify!($name), failed_attempts);
-                        errors.push(e);
                         return Err(Error::AllAttemptsErrored(errors));
                     }
 
-                    warn!("call '{}' failed with {}, retry: {}/{}", stringify!($name), e, failed_attempts, $self.config.retry());
-
-                    errors.push(e);
-
                     // Only one thread will try to recreate the client getting the write lock,
                     // other eventual threads will get Err and will block at the beginning of
                     // previous loop when trying to read()
@@ -80,15 +79,14 @@ macro_rules! impl_inner_call {
                                 Err(e) => {
                                     let failed_attempts = errors.len() + 1;
 
+                                    warn!("re-creating client failed with {}, retry: {}/{}", e, failed_attempts, $self.config.retry());
+
+                                    errors.push(e);
+
                                     if retries_exhausted(failed_attempts, $self.config.retry()) {
                                         warn!("re-creating client failed after {} attempts", failed_attempts);
-                                        errors.push(e);
                                         return Err(Error::AllAttemptsErrored(errors));
                                     }
-
-                                    warn!("re-creating client failed with {}, retry: {}/{}", e, failed_attempts, $self.config.retry());
-
-                                    errors.push(e);
                                 }
                             }
                         }
```

luisschwab:
@nabijaczleweli you need to run `just fmt` before pushing to save some back and forth.

PR #193, opened 2026-01-14
feat:tofu certificate validation

### LINKED ISSUE #176: feat:  check it out and use "TOFU" (trust on first use) for self-signed certificates
Instead of just ignoring and not validating the self-signed certificates, we probably should use the "TOFU" (trust on first use) for the self-signed certificates.

lucasdbr05:
Closes #176 

# What does this merge request do?

This feature (see the reference issue [https://github.com/bitcoindevkit/rust-electrum-client/issues/176](https://github.com/bitcoindevkit/rust-electrum-client/issues/176)) adds SSL certificate validation based on Trust On First Use (TOFU), storing the certificate on the first connection and verifying its consistency on subsequent connections.
It is implemented via the `TofuStore` trait, which allows customizable certificate persistence.

 - Introduces a refactored Trust On First Use (TOFU) implementation for SSL certificate verification.
- Adds the TofuStore trait, allowing library consumers to provide their own certificate persistence mechanism.
- Updates the `Client` configuration SSL client initialization flow to accept a TofuStore via ConfigBuilder.
- Adds examples in `examples` directory and scripts to demonstrate how TOFU can be integrated and used in practice.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,5 +1,7 @@
 //! Electrum Client
 
+#[cfg(feature = "tofu")]
+use std::sync::Arc;
 use std::{borrow::Borrow, sync::RwLock};
 
 use log::{info, warn};
@@ -10,6 +12,8 @@ use crate::api::ElectrumApi;
 use crate::batch::Batch;
 use crate::config::Config;
 use crate::raw_client::*;
+#[cfg(feature = "tofu")]
+use crate::tofu::TofuStore;
 use crate::types::*;
 use std::convert::TryFrom;
 
@@ -35,6 +39,8 @@ pub struct Client {
     client_type: RwLock<ClientType>,
     config: Config,
     url: String,
+    #[cfg(feature = "tofu")]
+    tofu_store: Option<Arc<dyn TofuStore>>,
 }
 
 macro_rules! impl_inner_call {
@@ -74,7 +80,7 @@ macro_rules! impl_inner_call {
                     if let Ok(mut write_client) = $self.client_type.try_write() {
                         loop {
                             std::thread::sleep(std::time::Duration::from_secs((1 << errors.len()).min(30) as u64));
-                            match ClientType::from_config(&$self.url, &$self.config) {
+                            match $self.client_type_adapter() {
                                 Ok(new_client) => {
                                     info!("Succesfully created new client");
                                     *write_client = new_client;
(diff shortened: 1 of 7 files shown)


src/tofu/mod.rs
```diff
@@ -0,0 +1,114 @@
+use std::fmt::{Debug};
+
+/// A trait for storing and retrieving TOFU (Trust On First Use) certificate data.
+/// Implementors of this trait are responsible for persisting certificate data and retrieving it based on the host.
+pub trait TofuStore: Send + Sync + Debug {
```
tnull: Why does this need to be `Send` + `Sync`?

tnull (commented):
Took an initial quick look and added some comments (though have yet to do a full review). Though before I proceed, please rewrite the commit history to avoid changing the approach mid-history. Basically, you should try to avoid to touch the same code paths in following commits as far as possible.

Given the code structure and verbosity I'm also suspecting that some form of AI agent was involved here. If this is indeed the case, please note that it is best practice to disclose such use in the PR and commit descriptions.

src/tofu/mod.rs
lucasdbr05: I did it this way to ensure that the Client remains thread-safe for concurrent applications, so I designed it to allow the store to be shared and safely accessed across multiple threads via an Arc.

src/raw_client.rs
```diff
@@ -446,24 +592,115 @@ impl RawClient<ElectrumSslStream> {
...
-        .map_err(Error::CouldNotCreateConnection)?;
+        .map_err(|e| {
+            let error_msg = format!("{}", e);
+            if error_msg.contains("TLS certificate changed") {
+                Error::TlsCertificateChanged(domain.clone())
+            } else if error_msg.contains("TOFU") {
+                Error::TofuPersistError(error_msg)
+            } else {
+                Error::CouldNotCreateConnection(e)
+            }
```
oleonardolima: It's not ideal to handle the error variants by containing a string on it, you should properly return an map a new error variant where needed.

oleonardolima (commented):
I agree with tnull's comments above, you should narrow it down to few commits, for example: feat(tofu): add tofu store mod; feat(client): add new tofu method/feature, docs(example): add new tofu example.

Also, it's best if it's added under a new feature, and by a separate constructor. I don't think many changes to already-existing methods are needed, maybe it's something remaining from previous changes you did. 

It's failing in CI, please make sure that everything is building successfully and passing CI too.

luisschwab:
Ok, this is finally in the top of my stack. @lucasdbr05 I'll start reviewing and testing this one.

src/tofu/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+        store: Mutex<HashMap<String, Vec<u8>>>,
+    }
+
+    impl InMemoryTofuStore {
+        fn new() -> Self {
+            Self {
+                store: Mutex::new(HashMap::new()),
+            }
+        }
+    }
```
oleonardolima: nit: you can probably get rid of this by using the `#[default]` macro as you did in the example.

src/client.rs
```diff
@@ -111,7 +114,11 @@ fn retries_exhausted(failed_attempts: usize, configured_retries: u8) -> bool {
 impl ClientType {
     /// Constructor that supports multiple backends and allows configuration through
     /// the [Config]
-    pub fn from_config(url: &str, config: &Config) -> Result<Self, Error> {
+    pub fn from_config(
+        url: &str,
+        config: &Config,
+        tofu_store: Option<Arc<dyn TofuStore>>,
```
oleonardolima: I think it's best to keep `ClientType::from_config` as it was, without a new parameter for the `tofu_store`, passing `None` when creating the client's below.

Instead, you can add a new method `from_config_with_tofu` for `ClientType` too.

src/client.rs
```diff
@@ -178,12 +199,34 @@ impl Client {
...
-        let client_type = RwLock::new(ClientType::from_config(url, &config)?);
+        let client_type = RwLock::new(ClientType::from_config(url, &config, None)?);
+
+        Ok(Client {
+            client_type,
+            config,
+            url: url.to_string(),
+            tofu_store: None,
+        })
+    }
```
oleonardolima: If you add the new `ClientType::from_config_with_tofu` you won't need these changes here.

src/client.rs
```diff
@@ -178,12 +199,34 @@ impl Client {
...
+    pub fn from_config_with_tofu(
+        url: &str,
+        config: Config,
+        tofu_store: Arc<dyn TofuStore>,
+    ) -> Result<Self, Error> {
+        let client_type = RwLock::new(ClientType::from_config(
+            url,
+            &config,
+            Some(tofu_store.clone()),
+        )?);
```
oleonardolima, line 286: You should use the new `ClienType::from_config_with_tofu` here.

src/lib.rs
```diff
@@ -81,3 +81,6 @@ pub use batch::Batch;
 pub use client::*;
 pub use config::{Config, ConfigBuilder, Socks5Config};
 pub use types::*;
+
+mod tofu;
```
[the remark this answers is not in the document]
oleonardolima: I agree, it's best to have it behind a feature, though I didn't try to check how messy it would become.

src/raw_client.rs
```diff
@@ -255,58 +258,105 @@ fn connect_with_total_timeout<A: ToSocketAddrs>(
...
-    /// `validate_domain` is `true`, validate the server's certificate.
+    /// Creates a new SSL client and tries to connect to `socket_addr`.
+    /// If `validate_domain` is `true`, validates the server's certificate.
+    /// If `tofu_store` is provided, uses TOFU (Trust On First Use) certificate validation, storing
+    /// the certificate on first connection and validating it matches on subsequent connections.
     pub fn new_ssl<A: ToSocketAddrsDomain + Clone>(
         socket_addrs: A,
         validate_domain: bool,
         timeout: Option<Duration>,
+        tofu_store: Option<Arc<dyn TofuStore>>,
```
oleonardolima: I think it's also best if we isolate this into a new method, such as: `new_ssl_with_tofu`, as it would make it easier to feature gate.

An even better alternative would be to use some way of builder pattern.

oleonardolima (commented):
I did an initial pass and left a few comments.

PR #194, opened 2026-01-16
feat!: add dynamic authorization support

EddieHouston:
### Description

This PR adds support for dynamic authorization in Electrum RPC requests. 
This enables authentication with authorization-protected Electrum servers and API gateways.

**Use cases:**
- Bearer token authentication (JWT, OAuth, API keys)
- API gateways requiring authorization headers
- Any dynamic authorization scenario where tokens need to be updated during the client's lifetime

### Notes to the reviewers

The implementation uses a callback pattern rather than a static token to support dynamic scenarios like automatic token refresh. The `AuthProvider` is called for each RPC request, allowing the token to be updated without reconnecting the client.

Both `Config` and `RawClient` structs needed custom `Debug` implementations since function pointers don't implement `Debug` - this is handled by displaying `<provider>` when an auth provider is present, which prevents leaking sensitive token values in debug output.

Thread safety is ensured through `Arc` wrapping and `Send + Sync` bounds, making this safe for concurrent use across the client.

**Tests added:**
    14 new tests, all passing

### Changelog notice

```
### Added

- Add new `AuthProvider` type alias (`Arc<dyn Fn() -> Option<String> + Send + Sync>`) for dynamic authorization token support (e.g. JWT Bearer, API keys, OAuth).

### Changed

- `ConfigBuilder` now accepts an optional `AuthProvider` via `authorization_provider()`, enabling token-refresh patterns without reconnecting.
- `RawClient` and `ClientType::from_config` updated to propagate the auth provider across all transport backends (TCP, SSL, SOCKS5).
```

#### All Submissions:

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

### DIFF
--- a/examples/jwt_auth.rs
+++ b/examples/jwt_auth.rs
@@ -1,0 +1,88 @@
+//! # JWT Static Authentication with Electrum Client
+//!
+//! This example demonstrates how to use a static JWT_TOKEN authentication with the
+//! electrum-client library.
+
+use bitcoin::Txid;
+use electrum_client::{Client, ConfigBuilder, ElectrumApi};
+use std::{str::FromStr, sync::Arc};
+
+const ELECTRUM_URL: &str = "ssl://electrum.blockstream.info:50002";
+
+const GENESIS_HEIGHT: usize = 0;
+const GENESIS_TXID: &str = "4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b";
+
+fn main() {
+    // A static JWT_TOKEN (i.e JWT_TOKEN="Bearer jwt_token...")
+    let auth_provider = Arc::new(move || {
+        let jwt_token = std::env::var("JWT_TOKEN").expect("JWT_TOKEN env variable not set");
+        Some(jwt_token)
+    });
+
+    // The Electrum Server URL (i.e `ELECTRUM_URL` environment variable, or defaults to `ELECTRUM_URL` const above)
+    let electrum_url = std::env::var("ELECTRUM_URL").unwrap_or(ELECTRUM_URL.to_owned());
+
+    // Builds the electrum-client `Config`.
+    let config = ConfigBuilder::new()
+        .validate_domain(false)
+        .authorization_provider(Some(auth_provider))
+        .build();
+
+    // Builds & Connect electrum-client `Client`.
+    match Client::from_config(&electrum_url, config) {
+        Ok(client) => {
+            println!(
+                "Successfully connected to Electrum Server: {:#?}; with JWT authentication!",
+                electrum_url
+            );
+
+            // try to call the `server.features` method, it can fail on some servers.
+            match client.server_features() {
+                Ok(features) => println!(
+                    "Successfully fetched the `server.features`!\n{:#?}",
+                    features
+                ),
+                Err(e) => eprintln!("Failed to fetch the `server.features`!\nError: {:#?}", e),
+            }
+
+            // try to call the `blockchain.block.header` method, it should NOT fail.
+            let genesis_height = GENESIS_HEIGHT;
+            match client.block_header(genesis_height) {
+                Ok(header) => {
+                    println!(
+                        "Successfully fetched the `Header` for given `height`={}!\n{:#?}",
+                        genesis_height, header
+                    );
+                }
+                Err(err) => eprintln!(
+                    "Failed to fetch the `Header` for given `height`!\nError: {:#?}",
+                    err
+                ),
+            }
+
+            // try to call the `blockchain.transaction.get` method, it should NOT fail.
+            let genesis_txid =
+                Txid::from_str(GENESIS_TXID).expect("SHOULD have a valid genesis `txid`");
+            match client.transaction_get(&genesis_txid) {
+                Ok(tx) => {
+                    println!(
+                        "Successfully fetched the `Transaction` for given `txid`={}!\n{:#?}",
+                        genesis_txid, tx
+                    );
+                }
+                Err(err) => eprintln!(
+                    "Failed to fetch the `Transaction` for given `txid`!\nError: {:#?}",
+                    err
+                ),
+            }
+        }
+        Err(err) => {
+            eprintln!(
+                "Failed to build and connect `Client` to {:#?}!\nError: {:#?}\n",
+                electrum_url, err
+            );
+            eprintln!("NOTE: This example requires an Electrum Server that handles/accept JWT authentication!");
+            eprintln!("Try to update the `ELECTRUM_URL` and `JWT_TOKEN to match your setup.");
+        }
+    }
+}
--- a/examples/jwt_dynamic_auth.rs
+++ b/examples/jwt_dynamic_auth.rs
@@ -1,0 +1,183 @@
+//! # JWT Dynamic Authentication
+//!
+//! ## Advanced: Token Refresh with Keycloak
+//!
+//! This example demonstrates how to use dynamic JWT authentication with the
+//! electrum-client library.
+//!
+//! ## Overview
+//!
+//! The electrum-client supports embedding authorization tokens (such as JWT
+//! Bearer tokens) directly in JSON-RPC requests. This is achieved through an
+//! [`AuthProvider`](electrum_client::config::AuthProvider) callback that is
+//! invoked before each request.
+//!
+//! In order to have an automatic token refresh (e.g it expires every 5 minutes),
+//! you should use a shared token holder (e.g KeycloakTokenManager)
+//! behind an `Arc<RwLock<...>>` and spawn a background task to refresh it.
+//!
+//! ## JSON-RPC Request Format
+//!
+//! With the auth provider configured, each JSON-RPC request will include the
+//! authorization field:
+//!
+//! ```json
+//! {
+//!   "jsonrpc": "2.0",
+//!   "method": "blockchain.headers.subscribe",
+//!   "params": [],
+//!   "id": 1,
+//!   "authorization": "Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..."
+//! }
+//! ```
+//!
+//! If the provider returns `None`, the authorization field is omitted from the
+//! request.
+//!
+//! ## Thread Safety
+//!
+//! The `AuthProvider` type is defined as:
+//!
+//! ```rust,ignore
+//! pub type AuthProvider = Arc<dyn Fn() -> Option<String> + Send + Sync>;
+//! ```
+//!
+//! This ensures thread-safe access to tokens across all RPC calls.
+
+use electrum_client::{Client, ConfigBuilder, ElectrumApi};
+use std::sync::{Arc, RwLock};
+use std::time::Duration;
+use tokio::time::sleep;
+
+/// Manages JWT tokens from Keycloak with automatic refresh
+struct KeycloakTokenManager {
+    token: Arc<RwLock<Option<String>>>,
+    keycloak_url: String,
+    grant_type: String,
+    client_id: String,
+    client_secret: String,
+}
+
+impl KeycloakTokenManager {
+    fn new(
+        keycloak_url: String,
+        grant_type: String,
+        client_id: String,
+        client_secret: String,
+    ) -> Self {
+        Self {
+            token: Arc::new(RwLock::new(None)),
+            keycloak_url,
+            client_id,
+            client_secret,
+            grant_type,
+        }
+    }
+
+    /// Get the current token (for the auth provider)
+    fn get_token(&self) -> Option<String> {
+        self.token.read().unwrap().clone()
+    }
+
+    /// Fetch a fresh token from Keycloak
+    async fn fetch_token(&self) -> Result<String, Box<dyn std::error::Error>> {
+        let url = format!("{}/protocol/openid-connect/token", self.keycloak_url);
+
+        // if you're using other HTTP client (i.e `reqwest`), you can probably use `.form` methods.
+        // it's currently not implemented in `bitreq`, needs to be built manually.
+        let body = format!(
+            "grant_type={}&client_id={}&client_secret={}",
+            self.grant_type, self.client_id, self.client_secret
+        );
+
+        let response = bitreq::post(url)
+            .with_header("Content-Type", "application/x-www-form-urlencoded")
+            .with_body(body)
+            .send_async()
+            .await?;
+
+        let json: serde_json::Value = response.json()?;
+        let access_token = json["access_token"]
+            .as_str()
+            .ok_or("Missing access_token")?
+            .to_string();
+
+        Ok(format!("Bearer {}", access_token))
+    }
+
+    /// Background task that refreshes the token every 4 minutes
+    async fn refresh_loop(self: Arc<Self>) {
+        loop {
+            // Refresh every 4 minutes (tokens expire at 5 minutes)
+            sleep(Duration::from_secs(240)).await;
+
+            match self.fetch_token().await {
+                Ok(new_token) => {
+                    println!("Token refreshed successfully");
+                    // In a background thread/task, periodically update the token
+                    *self.token.write().unwrap() = Some(new_token);
+                }
+                Err(e) => {
+                    eprintln!("Failed to refresh token: {}", e);
+                    // Keep using old token until we can refresh
+                }
+            }
+        }
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // The Electrum Server URL (i.e `ELECTRUM_URL` environment variable)
+    let electrum_url = std::env::var("ELECTRUM_URL")
+        .expect("SHOULD have the `ELECTRUM_URL` environment variable!");
+
+    // The JWT_TOKEN manager setup (i.e Keycloak server URL, client ID and secret)
+    let keycloak_url = std::env::var("KEYCLOAK_URL")
+        .expect("SHOULD have the `KEYCLOAK_URL` environment variable!");
+
+    let grant_type = std::env::var("GRANT_TYPE").unwrap_or("client_credentials".to_string());
+    let client_id =
+        std::env::var("CLIENT_ID").expect("SHOULD have the `CLIENT_ID` environment variable!");
+    let client_secret = std::env::var("CLIENT_SECRET")
+        .expect("SHOULD have the `CLIENT_SECRET` environment variable!");
+
+    // Setup `KeycloakTokenManager`
+    let token_manager = Arc::new(KeycloakTokenManager::new(
+        keycloak_url,
+        grant_type,
+        client_id,
+        client_secret,
+    ));
+
+    // Fetch initial token
+    let jwt_token = token_manager.fetch_token().await?;
+
+    println!("JWT_TOKEN='{}'", &jwt_token[..jwt_token.len().min(40)]);
+
+    *token_manager.token.write().unwrap() = Some(jwt_token);
+
+    // Start background refresh task
+    let tm_clone = token_manager.clone();
+    tokio::spawn(async move {
+        tm_clone.refresh_loop().await;
+    });
+
+    // Create Electrum client with dynamic auth provider
+    let tm_for_provider = token_manager.clone();
+    let config = ConfigBuilder::new()
+        .authorization_provider(Some(Arc::new(move || tm_for_provider.get_token())))
+        .build();
+
+    let client = Client::from_config(&electrum_url, config)?;
+
+    // All RPC calls will automatically include fresh JWT tokens
+    loop {
+        match client.server_features() {
+            Ok(features) => println!("Connected: {:?}", features),
+            Err(e) => eprintln!("Error: {}", e),
+        }
+
+        tokio::time::sleep(Duration::from_secs(10)).await;
+    }
+}
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,8 @@
+use std::sync::Arc;
 use std::time::Duration;
+
+/// A function that provides authorization tokens dynamically (e.g., for JWT refresh)
+pub type AuthProvider = Arc<dyn Fn() -> Option<String> + Send + Sync>;
 
 /// Configuration for an electrum client
 ///
@@ -6,7 +10,7 @@ use std::time::Duration;
 ///
 /// [`Client::from_config`]: crate::Client::from_config
 /// [`ClientType::from_config`]: crate::ClientType::from_config
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct Config {
     /// Proxy socks5 configuration, default None
     socks5: Option<Socks5Config>,
@@ -16,6 +20,24 @@ pub struct Config {
     retry: u8,
     /// when ssl, validate the domain, default true
     validate_domain: bool,
+    /// Optional authorization provider for dynamic token injection
+    authorization_provider: Option<AuthProvider>,
+}
+
+// Custom Debug impl because AuthProvider doesn't implement Debug
+impl std::fmt::Debug for Config {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Config")
+            .field("socks5", &self.socks5)
+            .field("timeout", &self.timeout)
+            .field("retry", &self.retry)
+            .field("validate_domain", &self.validate_domain)
+            .field(
+                "authorization_provider",
+                &self.authorization_provider.as_ref().map(|_| "<provider>"),
+            )
+            .finish()
+    }
 }
 
 /// Configuration for Socks5
@@ -69,6 +91,12 @@ impl ConfigBuilder {
     /// Sets if the domain has to be validated
     pub fn validate_domain(mut self, validate_domain: bool) -> Self {
         self.config.validate_domain = validate_domain;
+        self
+    }
+
+    /// Sets the authorization provider for dynamic token injection
+    pub fn authorization_provider(mut self, provider: Option<AuthProvider>) -> Self {
+        self.config.authorization_provider = provider;
         self
     }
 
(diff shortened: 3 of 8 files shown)


notmandatory:
concept ACK

Thanks for this, adding an "authorization" header is the most common way to authenticate to a http endpoints so this should be generally useful. Appreciate the comprehensive examples and unit tests. 

The team is spread a bit thin so it may take some time to get it reviewed and tested.

oleonardolima (commented):
cACK

I did an initial round of review. Overall, it looks good! I'll give it a try testing it.

EddieHouston:
> @EddieHouston needs rebase

@oleonardolima Was this still planned to release before the Electrum protocol v1.6 changes?

EddieHouston:
> > > @EddieHouston needs rebase
> > 
> > 
> > @oleonardolima Was this still planned to release before the Electrum protocol v1.6 changes?
> 
> I'm planning to have both under the same release, as the version negotiation is should work fine for new v1.6 and old servers. Appreciate it, if you could do a rebase.

Looks like changes for electrum protocol v1.6 break this.  The new negotiate_protocol_version are automatically called from constructors... those need to now pass auth as well so the call to server.version can succeed.  
Update coming soon

EddieHouston:
> > > > @EddieHouston needs rebase
> > > 
> > > 
> > > @oleonardolima Was this still planned to release before the Electrum protocol v1.6 changes?
> > 
> > 
> > I'm planning to have both under the same release, as the version negotiation is should work fine for new v1.6 and old servers. Appreciate it, if you could do a rebase.
> 
> Looks like changes for electrum protocol v1.6 break this. The new negotiate_protocol_version are automatically called from constructors... those need to now pass auth as well so the call to server.version can succeed. Update coming soon

I asked Claude Code to fix this issue (squashed the change and force pushed):

 Here's the old flow:                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  RawClient::new_ssl(url, validate_domain, timeout)   // 1. creates stream                                                                                                                                                                          
                                                        // 2. negotiate_protocol_version() ← sends server.version WITHOUT auth
                                                        // 3. returns client
      .with_auth_provider(config.auth_provider())       // 4. sets auth ← TOO LATE

  The electrum server requiring auth also requires auth on server.version, so step 2 fails before we ever get to step 4.

  The _with_auth constructors fix the ordering:

  RawClient::new_ssl_with_auth(url, validate_domain, timeout, auth_provider)
      // 1. creates stream
      // 2. client.auth_provider = auth_provider  ← auth set FIRST
      // 3. negotiate_protocol_version()          ← sends server.version WITH auth ✓


Tested locally and this works.

oleonardolima:
> I asked Claude Code to fix this issue (squashed the change and force pushed):
> 
> Here's the old flow:
> 
> RawClient::new_ssl(url, validate_domain, timeout) // 1. creates stream // 2. negotiate_protocol_version() ← sends server.version WITHOUT auth // 3. returns client .with_auth_provider(config.auth_provider()) // 4. sets auth ← TOO LATE
> 
> The electrum server requiring auth also requires auth on server.version, so step 2 fails before we ever get to step 4.
> 
> The _with_auth constructors fix the ordering:
> 
> RawClient::new_ssl_with_auth(url, validate_domain, timeout, auth_provider) // 1. creates stream // 2. client.auth_provider = auth_provider ← auth set FIRST // 3. negotiate_protocol_version() ← sends server.version WITH auth ✓
> 
> Tested locally and this works.

@EddieHouston It looks like Claude didn't fully fixed the issues, CI is failing.

oleonardolima:
> > I asked Claude Code to fix this issue (squashed the change and force pushed):
> > Here's the old flow:
> > RawClient::new_ssl(url, validate_domain, timeout) // 1. creates stream // 2. negotiate_protocol_version() ← sends server.version WITHOUT auth // 3. returns client .with_auth_provider(config.auth_provider()) // 4. sets auth ← TOO LATE
> > The electrum server requiring auth also requires auth on server.version, so step 2 fails before we ever get to step 4.
> > The _with_auth constructors fix the ordering:
> > RawClient::new_ssl_with_auth(url, validate_domain, timeout, auth_provider) // 1. creates stream // 2. client.auth_provider = auth_provider ← auth set FIRST // 3. negotiate_protocol_version() ← sends server.version WITH auth ✓
> > Tested locally and this works.
> 
> @EddieHouston It looks like Claude didn't fully fixed the issues, CI is failing.

I asked it to fix it, and it gave this suggestions, see: 17fc5b6522f698acf19ef24fef111361b4634d4d

Oh, still failing though 🤦

EddieHouston:
> > I asked Claude Code to fix this issue (squashed the change and force pushed):
> > Here's the old flow:
> > RawClient::new_ssl(url, validate_domain, timeout) // 1. creates stream // 2. negotiate_protocol_version() ← sends server.version WITHOUT auth // 3. returns client .with_auth_provider(config.auth_provider()) // 4. sets auth ← TOO LATE
> > The electrum server requiring auth also requires auth on server.version, so step 2 fails before we ever get to step 4.
> > The _with_auth constructors fix the ordering:
> > RawClient::new_ssl_with_auth(url, validate_domain, timeout, auth_provider) // 1. creates stream // 2. client.auth_provider = auth_provider ← auth set FIRST // 3. negotiate_protocol_version() ← sends server.version WITH auth ✓
> > Tested locally and this works.
> 
> @EddieHouston It looks like Claude didn't fully fixed the issues, CI is failing.

The CI runs with --no-default-features --features proxy (no SSL). 
The new_proxy_ssl_with_auth method references ElectrumSslStream but is missing the #[cfg] gate that the public new_proxy_ssl has.

@oleonardolima Updated with fix.  Ran the `cargo check --no-default-features --features proxy` locally and see it works now.

EddieHouston:
@oleonardolima I'll ran these locally and resolved the issues:
   cargo clippy --all-features --all-targets -- -D warnings
   cargo fmt --all -- --config format_code_in_doc_comments=true --check
   cargo test --verbose --all-features
   
Newbee question...   Is there an easy way to run all the CI test locally?

EddieHouston:
> @oleonardolima I'll ran these locally and resolved the issues: cargo clippy --all-features --all-targets -- -D warnings cargo fmt --all -- --config format_code_in_doc_comments=true --check cargo test --verbose --all-features
> 
> Newbee question... Is there an easy way to run all the CI test locally?
Figured it out...

 # 1. fmt                                                                                                                                                                                                                                          
  cargo fmt --all -- --config format_code_in_doc_comments=true --check                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  # 2. clippy                                                                                                                                                                                                                                       
  cargo clippy --all-features --all-targets -- -D warnings                                                                                                                                                                                          

  # 3. tests
  cargo test --verbose --all-features

  # 4. check all feature combinations
  cargo check --verbose --features default
  cargo check --verbose --no-default-features
  cargo check --verbose --no-default-features --features proxy
  cargo check --verbose --no-default-features --features openssl
  cargo check --verbose --no-default-features --features rustls
  cargo check --verbose --no-default-features --features rustls-ring
  cargo check --verbose --no-default-features --features proxy,openssl,rustls,rustls-ring

  # 5. 
  rustup toolchain install 1.75.0
  cargo +1.75.0 check --verbose --all-features

src/types.rs
```diff
@@ -518,11 +522,77 @@ impl From<std::sync::mpsc::RecvError> for Error {
...
+
+    #[test]
+    fn test_authorization_field_omitted_when_none() {
+        let req = Request::new_id(1, "test.method", vec![]);
+
+        let json = serde_json::to_string(&req).unwrap();
+
+        // The JSON should not contain the word "authorization" at all
+        assert!(!json.contains("authorization"));
+    }
```
oleonardolima: nit: can't you could just move this assertion to the first test ?

src/raw_client.rs
```diff
@@ -159,9 +159,32 @@ where
     /// The protocol version negotiated with the server via `server.version`.
     protocol_version: Mutex<Option<String>>,
 
+    auth_provider: Option<AuthProvider>,
```
oleonardolima, line 163: nit: could use documentation

src/raw_client.rs
```diff
@@ -568,7 +696,7 @@ impl<S: Read + Write> RawClient<S> {
     ///
     /// This sends `server.version` as the first message and stores the negotiated
     /// protocol version. Called automatically by constructors.
-    fn negotiate_protocol_version(&self) -> Result<(), Error> {
+    pub(crate) fn negotiate_protocol_version(&self) -> Result<(), Error> {
```
oleonardolima: question: what's the rationale on making this pub(crate) ? I didn't see any new usage of it.

oleonardolima (commented):
utACK 09345de0311c70e78e41f5d95df5274e30a39d0c

It looks good, though I'm wondering what's the rationale on adding new methods, instead of update the existing ones APIs. Did you want to reduce the number of breaking changes ?

I'll come up with a patch with some changes I have in mind, to see if the API would be better.

src/raw_client.rs
EddieHouston: The pub(crate) was left over from an intermediate debugging approach. Reverted to fn.

src/types.rs
EddieHouston: Sure thing 👍

src/raw_client.rs
EddieHouston: Added

EddieHouston:
> utACK [09345de](https://github.com/bitcoindevkit/rust-electrum-client/commit/09345de0311c70e78e41f5d95df5274e30a39d0c)
> 
> It looks good, though I'm wondering what's the rationale on adding new methods, instead of update the existing ones APIs. Did you want to reduce the number of breaking changes ?
> 
> I'll come up with a patch with some changes I have in mind, to see if the API would be better.

Yes, exactly... avoiding breaking changes for existing clients.

examples/JWT_AUTH_EXAMPLE.md
```diff
@@ -0,0 +1,196 @@
...
+If the provider returns `None`, the authorization field is omitted from the request.
+
+## Thread Safety
+
+The `AuthProvider` type is defined as:
+```rust
+pub type AuthProvider = Arc<dyn Fn() -> Option<String> + Send + Sync>;
+```
+
+This ensures thread-safe access to tokens across all RPC calls.
```
luisschwab: You can remove this `JWT_AUTH_EXAMPLE.md` and add it to `jwt_auth.rs` as a doc comment so that it shows up on docs.rs and removes some duplication.

luisschwab (changes requested):

examples/JWT_AUTH_EXAMPLE.md
EddieHouston: Good call, done.

oleonardolima (commented):
tACK 020f193ac84f70af977b8482e9424fec57ef34cd

It's looking good! I still think the API feels a bit odd, and we could just update the existing methods to have a new auth parameter, though it can be addressed in a follow-up (maybe in another release even).

I gave it a shot and tested it with non-authenticated (testnet) servers that support v1.6 by syncing/broadcasting w/ bdk_wallet, it's still working as expected.

I also tested it with authenticated (mainnet) servers, but didn't tested it by syncing/broadcasting w/ bdk_wallet, still need to test it in testnet/signet if possible.

src/config.rs
```diff
@@ -131,6 +159,13 @@ impl Config {
         self.validate_domain
     }
 
+    /// Get the configuration for `authorization_provider`
+    ///
+    /// Set this with [`ConfigBuilder::authorization_provider`]
+    pub fn authorization_provider(&self) -> &Option<AuthProvider> {
+        &self.authorization_provider
+    }
+
```
luisschwab, line 168: Return an `Option` to a reference so callers don't have to match on a reference to an `Option`.

```suggestion
    /// Get the configuration for `authorization_provider`
    ///
    /// Set this with [`ConfigBuilder::authorization_provider`]
    pub fn authorization_provider(&self) -> Option<&AuthProvider> {
        self.authorization_provider.as_ref()
    }

```

src/raw_client.rs
```diff
@@ -832,12 +967,20 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
                 method,
                 params.to_vec(),
             );
+
+            // Apply authorization token only to the first request in batch
+            if index == 0 {
+                if let Some(provider) = &self.auth_provider {
+                    req.authorization = provider();
+                }
+            }
```
luisschwab: @EddieHouston is it the case that all servers expect the auth token in the first request only in a batched request? I haven't tested this out personally.

src/raw_client.rs
```diff
@@ -553,6 +675,13 @@ enum ChannelMessage {
 }
 
 impl<S: Read + Write> RawClient<S> {
+    /// Sets the authorization provider for dynamic token injection.
+    /// This should be called immediately after constructing the client.
+    pub fn with_auth_provider(mut self, provider: Option<AuthProvider>) -> Self {
+        self.auth_provider = provider;
+        self
+    }
+
```
luisschwab: This shouldn't be `pub`, since it's a workaround for adding a provider after building a test client with `get_test_client()`. You could add a `get_test_client_with_auth(provider: AuthProvider)` and remove this.

```rs
fn get_test_client_with_auth(provider: AuthProvider) -> RawClient<ElectrumSslStream> {
    let server =
        std::env::var("TEST_ELECTRUM_SERVER").unwrap_or("fortress.qtornado.com:443".into());
    RawClient::new_ssl_with_auth(&*server, false, None, Some(provider)).unwrap()
}
```

luisschwab (changes requested):

examples/jwt_auth.rs
```diff
@@ -0,0 +1,115 @@
...
+//! // In a background thread/task, periodically update the token:
+//! // *token.write().unwrap() = Some("Bearer refreshed-token".to_string());
+//! ```
+//!
+//! ## Integration with BDK
+//!
+//! To use with BDK, create the electrum client with your config, then wrap it:
+//!
+//! ```rust,ignore
+//! use bdk_electrum::BdkElectrumClient;
```
luisschwab:
```suggestion
```

examples/jwt_auth.rs
```diff
@@ -0,0 +1,115 @@
...
+//!
+//! let config = ConfigBuilder::new()
+//!     .authorization_provider(Some(Arc::new(move || {
+//!         token_manager.get_token()
+//!     })))
+//!     .timeout(Some(Duration::from_secs(30)))
+//!     .build();
+//!
+//! let electrum_client = Client::from_config("ssl://your-api-gateway:50002", config)?;
+//! let bdk_client = BdkElectrumClient::new(electrum_client);
```
luisschwab:
```suggestion
```

luisschwab (dismissed):
These should be removed, since `electrum-client` is unaware of `bdk_electrum`. This is a dependency loop.

src/raw_client.rs
EddieHouston: yes, thats correct.

examples/jwt_auth.rs
EddieHouston: just removing those few lines defeats the purpose of showing how to use this client change in bdk. 
We might as well remove the whole section if the helper example isn't appropriate here.

src/raw_client.rs
luisschwab: Can you explain that in the comment?

luisschwab:
> I still think the API feels a bit odd, and we could just update the existing methods to have a new auth parameter, though it can be addressed in a follow-up (maybe in another release even).

@EddieHouston how hard would it be run this in regtest on CI? Since we'll probably move some stuff around for the next release, would be nice to be able to test it out there and catch any regressions.

luisschwab:
@EddieHouston can you also provide complete instructions on testing this with your servers?

oleonardolima:
> @EddieHouston can you also provide complete instructions on testing this with your servers?

@luisschwab You can probably start from this: https://github.com/oleonardolima/rust-electrum-client/commit/4f6bc16a3ba398d6a5676208e14aa2b1fa7137a2

I think that having it as a full example is better than have it all in just docs, will update to the latest changes and probably push that commit, wdyt @EddieHouston

luisschwab:
@oleonardolima go ahead. While you're at it, can you squash the last 3 commits and fix formatting?

EddieHouston:
> > I still think the API feels a bit odd, and we could just update the existing methods to have a new auth parameter, though it can be addressed in a follow-up (maybe in another release even).
> 
> @EddieHouston how hard would it be run this in regtest on CI? Since we'll probably move some stuff around for the next release, would be nice to be able to test it out there and catch any regressions.

In theory not difficult, just time.  You could maybe mock up a test server that takes an API key or something.

oleonardolima:
I cherry-picked my changes for the examples and squashed everything into a single commit.

I changed the `jwt_dynamic_auth` example to use `bitreq` instead of `reqwest`, as `reqwest` was bringing a bunch of MSRV issues on CI.

Adding a regtest auth infra for testing in CI can be done in a follow-up.

oleonardolima (approved):
tACK 1c7ad73c61399c444c8af5d30e8ff3c8de4d3291

I've tested using this version of the `electrum-client` for `FullScan`/`Sync` and broadcasting with a `bdk_wallet` on `Testnet3`. I did it for both unauthenticated and authenticated servers, and it's working as expected.

src/config.rs
```diff
@@ -1,12 +1,16 @@
+use std::sync::Arc;
 use std::time::Duration;
 
+/// A function that provides authorization tokens dynamically (e.g., for JWT refresh)
+pub type AuthProvider = Arc<dyn Fn() -> Option<String> + Send + Sync>;
```
evanlinjin, line 5: The `config` submodule is private - let's reexport this in `lib.rs`.

src/raw_client.rs
```diff
@@ -832,12 +960,20 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
-        for (method, params) in batch.iter() {
-            let req = Request::new_id(
+        for (index, (method, params)) in batch.iter().enumerate() {
+            let mut req = Request::new_id(
                 self.last_id.fetch_add(1, Ordering::SeqCst),
                 method,
                 params.to_vec(),
             );
+
+            // Servers expect auth token only in the first request in batched requests
```
evanlinjin: Unfortunately this library doesn't actually use JSON-RPC batch arrays - the "batch" here just means "write all in one go".

We need to add auth token to all requests.

src/types.rs
```diff
@@ -71,6 +71,9 @@ pub struct Request<'a> {
     pub method: &'a str,
     /// The request parameters
     pub params: Vec<Param>,
+    /// Optional authorization token (e.g., JWT Bearer token)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub authorization: Option<String>,
```
evanlinjin: Is this intentionally made public? `RawClient` seems to overwrite this regardless (`raw_client.rs:973`).

Either make this `pub(crate)`, or change logic to not override a caller-set value (however, does this feature even make sense?).

src/raw_client.rs
```diff
@@ -196,6 +222,14 @@ impl RawClient<ElectrumPlaintextStream> {
     pub fn new<A: ToSocketAddrs>(
         socket_addrs: A,
         timeout: Option<Duration>,
+    ) -> Result<Self, Error> {
+        Self::new_with_auth(socket_addrs, timeout, None)
+    }
+
+    pub(crate) fn new_with_auth<A: ToSocketAddrs>(
```
evanlinjin: All of the new `.._with_auth` methods seem to only populate the `auto_provider` field. Is this correct?

Instead of having all these duplicate methods - we can use a builder pattern by introducing a single method:

```rust
pub(crate) with_auth(mut self, auth_provider: AuthProvider) -> Self {
    self.auth_provider = Some(auth_provider);
    self
}
```

evanlinjin (dismissed):
Thanks for this work. This is only a partial review.

src/types.rs
oleonardolima: If I got it right, `RawClient` overwrites it because `Request` is always being created with an `authorization` as `None`.

Yes, it probably can be `pub(crate)` and already assigned when creating a new `Request`, that would also fix the issue of only adding the authorization to the first one in batch scenarios.

src/raw_client.rs
oleonardolima: @EddieHouston Are there any issues on the server side if we choose to send the authorization within every request ?

src/config.rs
oleonardolima: I fixed this in 72dc88a37cdc86bc47d91337d1edf1eb16879b14

src/raw_client.rs
oleonardolima: Anyway, I fixed this in 72dc88a37cdc86bc47d91337d1edf1eb16879b14

src/types.rs
oleonardolima: I fixed this in 72dc88a37cdc86bc47d91337d1edf1eb16879b14

src/raw_client.rs
oleonardolima: I think this goes back to my previous [comment](https://github.com/bitcoindevkit/rust-electrum-client/pull/194#pullrequestreview-3900422503).

Yes, we should give it a try to builder pattern, it'll give us a better API.
cc @EddieHouston

src/raw_client.rs
EddieHouston: > @EddieHouston Are there any issues on the server side if we choose to send the authorization within every request ?

yeah, it's a huge amount of extra unneeded data over the wire.

oleonardolima: > > @EddieHouston Are there any issues on the server side if we choose to send the authorization within every request ?
> 
> yeah, it's a huge amount of extra unneeded data over the wire.

Oh, got it. However, we should probably follow with above's suggestion to guarantee general compatibility.

EddieHouston: I don't see the problem, the original code adds the auth only on the first request in the batch, even though its using newline-delimited JSON-RPC, they are all flushed to the wire together and it has auth only on the first one as expected.

oleonardolima: I think we can keep as it's for now, if leads to any performance issues we can change it and release in a patch release, as it'd be internal and non-breaking.

src/types.rs
oleonardolima: Actually I changed it to private in the latest changes, there's no need to be either pub or pub(crate), you should set it through the new `.with_auth()` method.

src/raw_client.rs
oleonardolima: Alright, I did it in the latest commits. It now has a better API and follows the builder pattern.

src/raw_client.rs
oleonardolima: > I think we can keep as it's for now, if leads to any performance issues we can change it and release in a patch release, as it'd be internal and non-breaking.

I updated it to use this argument in favor of only adding in the first request, if there's any other user with this usecase, which require the auth token in every request we can do it in the future without a breaking change.

oleonardolima:
I added two new commits that addresses:

(i) the protocol version negotiation being strictly called internally, instead of leaving it up to the user (as discussed with tnull in discord).

(ii) the `auth_provider` is now passed to the `RawClient` methods, as it's required to be available prior to protocol negotiation.

(iii) applying the authorization token strictly in the first request of a batch call, as mentioned by @EddieHouston [here](https://github.com/bitcoindevkit/rust-electrum-client/pull/194#discussion_r2945180510).

I kept it as separate commits to make clear what the changes are. 

--

It's now ready again for a new round of review.
cc @luisschwab @notmandatory

examples/jwt_auth.rs
```diff
@@ -0,0 +1,88 @@
+//! # JWT Static Authentication with Electrum Client
+//!
+//! This example demonstrates how to use an static JWT_TOKEN authentication with the
```
EddieHouston: Nit: "an static" → "a static"

src/raw_client.rs
```diff
@@ -832,12 +909,25 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
...
-            let req = Request::new_id(
+        for (idx, (method, params)) in batch.iter().enumerate() {
+            let mut req = Request::new_id(
                 self.last_id.fetch_add(1, Ordering::SeqCst),
                 method,
                 params.to_vec(),
             );
+
+            // although the library DOES NOT use JSON-RPC batch arrays,
+            // it applies the `authorization` ONLY in the first `Request` of the `Batch`.
```
EddieHouston: Worth noting in the comment why:

"JWTs can be 1KB+ and duplicating across 100 requests adds significant overhead.  This assumes the server authenticates the batch from the first request. If a server implementation treats each newline-delimited request independently, requests 2+ would be unauthenticated. 
This is a known trade-off, not a bug."

examples/jwt_auth.rs
```diff
@@ -0,0 +1,88 @@
...
+                    err
+                ),
+            }
+        }
+        Err(err) => {
+            eprintln!(
+                "Failed to build and connect `Client` to {:#?}!\nError: {:#?}\n",
+                electrum_url, err
+            );
+            eprintln!("NOTE: This example requires an Electrum Server that handle/accept JWT authentication!");
```
EddieHouston: Nit: "handle" → "handles"

examples/jwt_dynamic_auth.rs
```diff
@@ -0,0 +1,183 @@
...
+        keycloak_url,
+        grant_type,
+        client_id,
+        client_secret,
+    ));
+
+    // Fetch initial token
+    let jwt_token = token_manager.fetch_token().await?;
+
+    println!("JWT_TOKEN='{}'", &jwt_token);
```
EddieHouston: This logs the full bearer token. Since examples get copy-pasted, this could lead to token leakage in production logs. Consider truncating:                                                                                                      
  `println!("JWT_TOKEN='{}'", &jwt_token[..jwt_token.len().min(40)]);     `

EddieHouston (commented):
Overall great improvements, thanks!  
On review found one small change in logging the JWT, a small comment, and a couple nits.

oleonardolima:
> Overall great improvements, thanks! On review found one small change in logging the JWT, a small comment, and a couple nits.

Good, thanks for the review! I addressed your comments.

oleonardolima (approved):
tACK 9821dc85c24489ae70bc88aab988b54de677cc71

I tested it again in `testnet3` by scanning/syncing a `bdk_wallet` with both:

- unauthenticated server (v1.6): building a zero-fee tx (V3), broadcasting it (got `min relay fee` expected error), and was able to broadcast it with a transaction package.

- authenticated server (v1.4): building a tx (V2), broadcasting it, and then doing a RBF.

luisschwab (approved):
tACK 9821dc85c24489ae70bc88aab988b54de677cc71

Issue #166, opened 2025-03-07
Allow to construct `ElectrumClient` *without* connecting


tnull:
Currently, `ElectrumClient::new` directly tries to establish a connection to the server, i.e., there is no way to successfully construct a client while being offline. 

It would be great if there was a way to simply initialize a client, without directly establishing a connection.

This might go hand-in-hand with decoupling the `Client` object more generally from a single connection, see #165.

reez:
A similar issue https://github.com/bitcoindevkit/rust-electrum-client/issues/212 was opened related to this, so I'll close that and use this pre-existing issue instead. Noting the comments on it as reference:

https://github.com/bitcoindevkit/rust-electrum-client/issues/212#issue-4729125264

https://github.com/bitcoindevkit/rust-electrum-client/issues/212#issuecomment-4783785225

reez:
I opened a PR to explore addressing this issue and 212

PR #214, opened 2026-06-25
feat(client): allow constructing without connecting

reez:
#### Description

Attempts to address https://github.com/bitcoindevkit/rust-electrum-client/issues/166

Allows `Client` to be constructed without opening a network connection, adds `Client::connect()` for explicit connection and protocol negotiation, and defers automatic connection until the first Electrum API call.

#### Notes to Reviewers

I kept the patch intentionally small so I left out things like:
- Adding an eager constructor like `from_config_connected` for callers who want the previous construct-and-connect behavior in one step.
- Refactoring the duplicated `ssl://` no-TLS check into a small helper if that feels cleaner.

#### Related Issues

This started for me from [bdk-dart#91](https://github.com/bitcoindevkit/bdk-dart/issues/91), where constructing an FFI backed `ElectrumClient` can block the Android UI thread during TCP connection and cause ANRs.

I opened [bdk-ffi#1037](https://github.com/bitcoindevkit/bdk-ffi/issues/1037) to address it on bdk-ffi, trying to make that blocking boundary clearer in the bindings, but I think a bdk-ffi only change still falls a bit short without an upstream change from `rust-electrum-client`.

So this PR is for that upstream client layer change, and is closely related to [rust-electrum-client#166](https://github.com/bitcoindevkit/rust-electrum-client/issues/166) where tnull asked for a way to construct an Electrum client without directly establishing a connection.


luisschwab:
@oleonardolima you can drop the `zeroize` pin from this PR if you merge #217 first.

caarloshenriq (approved):
tACK 3d03a44

Built and ran the test suite across the feature matrix (default, `rustls-ring,proxy`, `proxy` only) plus `cargo clippy` on rustc 1.85, all clean.

Verified the eager `ssl://` check fires correctly with no TLS feature enabled, confirmed via `cargo tree` that no TLS backend was in the resolved feature set.

PR #218, opened 2026-07-02
fix: redact authorization tokens from trace logs

EddieHouston:
fixes #215.

### Description

The `authorization` token added via `AuthProvider` (#194) was logged verbatim
at TRACE level in `call` and `batch_call`, since both traced the raw wire
bytes.

This redacts the token before serialization instead of inspecting serialized
bytes:

- new crate-private `Request::redacted` returns a copy of the request with the
  token replaced by `<REDACTED>`
- `call` traces the redacted request; `batch_call` collects its requests into
  a `Vec` and derives both the wire bytes and the trace output from it

### Notes to the reviewers

Redacting the struct before serialization removes the JSON re-parse and the
fail-closed path of the previous approach, and keeps the trace line in the
same field order as the wire format. `batch_call` still attaches the token
only to the first request of the batch.

### Changelog notice

```
### Fixed

- Fix `authorization` token leaking into TRACE logs in `call` and `batch_call` #218

```

#### All Submissions:

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [x] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

### DIFF
--- a/src/raw_client.rs
+++ b/src/raw_client.rs
@@ -912,7 +912,7 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
     }
 
     fn batch_call(&self, batch: &Batch) -> Result<Vec<serde_json::Value>, Error> {
-        let mut raw = Vec::new();
+        let mut requests = Vec::new();
 
         let mut missing_responses = Vec::new();
         let mut answers = BTreeMap::new();
@@ -951,15 +951,27 @@ impl<T: Read + Write> ElectrumApi for RawClient<T> {
 
             self.waiting_map.lock()?.insert(req.id, sender);
 
-            raw.append(&mut serde_json::to_vec(&req)?);
-            raw.extend_from_slice(b"\n");
+            requests.push(req);
         }
 
         if missing_responses.is_empty() {
             return Ok(vec![]);
         }
 
-        trace!("==> {}", String::from_utf8_lossy(&raw));
+        let mut raw = Vec::new();
+        for req in &requests {
+            raw.append(&mut serde_json::to_vec(req)?);
+            raw.extend_from_slice(b"\n");
+        }
+
+        trace!(
+            "==> {}",
+            requests
+                .iter()
+                .map(|req| serde_json::to_string(&req.redacted()))
+                .collect::<Result<Vec<String>, _>>()?
+                .join("\n")
+        );
 
         let mut stream = self.stream.lock()?;
         stream.write_all(&raw)?;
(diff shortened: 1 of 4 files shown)


EddieHouston:
note... had to pin zeroize in new commit due to MSRV:  
image

src/raw_client.rs
```diff
@@ -1439,6 +1484,46 @@ mod test {
             .expect("should build the `RawClient` successfully!")
     }
 
+    #[test]
+    fn test_redact_authorization() {
+        let raw = br#"{"jsonrpc":"2.0","id":1,"method":"server.version","params":[],"authorization":"Bearer secret-token"}"#;
+        let redacted = redact_authorization(raw);
+        let request: serde_json::Value = serde_json::from_slice(&redacted).unwrap();
+
+        assert_eq!(request["authorization"], "<redacted>");
```
sdmg15: nit: I think we can use the constant value `REDACTED_AUTHORIZATION` here

sdmg15 (commented):
tAck ab750e3b5110c32507e278bfc747d421aa83f0f8.

I also tested a case in batch requests where one is valid json and another is not and in the logs we have `==> <request redaction failed>`

src/raw_client.rs
EddieHouston: Good catch, done. Both test assertions now use REDACTED_AUTHORIZATION.

oleonardolima:
@EddieHouston I did pushed a commit to fix the CI, and some others with a different approach. Let me know what you think.

I don't want this one to be a breaking change, so we can have it as patch release. However, I think the ideal solution would be to have this constraint at type level as follow-up.

oleonardolima:
ping @luisschwab

luisschwab:
@oleonardolima missing the changelog entry

EddieHouston:
> @EddieHouston I did pushed a commit to fix the CI, and some others with a different approach. Let me know what you think.
> 
> I don't want this one to be a breaking change, so we can have it as patch release. However, I think the ideal solution would be to have this constraint at type level as follow-up.

Fine with me, as long as it builds... pinning whatever needs to is great 👍

sdmg15 (approved):
re-tAck https://github.com/bitcoindevkit/rust-electrum-client/pull/218/changes/77f9c784f461392024aafb67ae96b55062434431

oleonardolima (approved):
self-ACK 77f9c784f461392024aafb67ae96b55062434431

